PR #14, opened 2013-04-26
[NO_MERGE] [JENKINS-16316] - Bug fix and additional tests

kazesberger:
slave (which are outdated if you dont restart slave nodes after changing
global properties)

CAUTION - this might break some features: EnvironmentContributors don't
get recognized any more (because CoreEnvironmentContributor did a
Computer.current.getEnvironment() which also leads to overwriting with
outdated values.

how to reproduce the problem:
hpi:run on current head with jenkins-version set to "1.480.3" (current LTS verison)
configure a global property on master K/V: TESTVAR/value1
configure a slave and start it
configure a job restricted to run on that slave and to echo $TESTVAR.
run the job (it will echo "value1")
configure the global property on master to be: TESTVAR/value2
dont restart anything
rerun the job (it will still echo "value1")
however a job on master will recognize the changed value.
there are 2 pieces of code that lead to this behaviour. I'm pretty uncertain about the fix i made, but it resolves at least our problem with the plugin as i described it above, although i'm pretty sure it breaks some other intended feature.
https://github.com/kazesberger/envinject-plugin.git
https://github.com/kazesberger/envinject-plugin/commits/master
maybe someone can have a look at it and/or tell me what's the story behind this (Core)EnvironmentContributor thing.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,5 +1,5 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
 
     <modelVersion>4.0.0</modelVersion>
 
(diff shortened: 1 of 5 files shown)


recena:
@kazesberger Do you agree if I close this PR?

oleg-nenashev:
@recena 
Please don't rush. The PR contains a useful test, which may be decoupled. That's why we keep this PR open

recena:
@oleg-nenashev Ok, Do you agree if we move the interested tests to a new PR and close this?

oleg-nenashev:
@recena 
I definitely agree. I was just too lazy to do it, so I kept the PR :)

recena:
@oleg-nenashev I'll do it.

PR #20, opened 2013-07-26
[JENKINS-16316] Changes to global variables not honored.

jglick:
Seems to fix [JENKINS-16316](https://issues.jenkins-ci.org/browse/JENKINS-16316). When `EnvInjectComputerListener.onOnline` considers `Jenkins.globalNodeProperties`, it takes the initial values of these variables and stuffs them into `masterEnvVars` on the slave, which really ought to be limited to “natural” environment variables. Then when these variables are changed and a build runs, the old values from `masterEnvVars` take precedence. This is true even if the user has never done anything with the EnvInject plugin except install it—especially insidious.

Pull #14 also claims to fix the same bug, in a very different way. Even if that is rejected, its tests may be useful.


jglick:
Could be. Use `git bisect` to see. If so, mark JENKINS-18297 as blocking JENKINS-16316. Obviously more complete functional tests are needed for this plugin.

PR #36, opened 2014-10-20
JENKINS-25223: Allow executing scripts in JobProperty.prebuild phase

arcivanov:
Creates the "prebuild" phase JobProperty section, allowing scripts to be executed after SCM is completed but before AbstractBuildExecution.doRun kicks in.
Also fixes NPE in EnvInjectJobProperty where disabling a section and saving before applying causes NPE on save

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPrebuildJobProperty.java
+++ b/src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPrebuildJobProperty.java
@@ -1,0 +1,390 @@
+package org.jenkinsci.plugins.envinject;
+
+import hudson.DescriptorExtensionList;
+import hudson.Extension;
+import hudson.FilePath;
+import hudson.model.BuildListener;
+import hudson.model.Environment;
+import hudson.model.JobProperty;
+import hudson.model.JobPropertyDescriptor;
+import hudson.model.AbstractBuild;
+import hudson.model.Computer;
+import hudson.model.Descriptor;
+import hudson.model.Job;
+import hudson.model.Node;
+import hudson.model.Result;
+import hudson.model.Run;
+
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+import javax.annotation.CheckForNull;
+
+import jenkins.model.Jenkins;
+import net.sf.json.JSON;
+import net.sf.json.JSONException;
+import net.sf.json.JSONObject;
+
+import org.jenkinsci.lib.envinject.EnvInjectException;
+import org.jenkinsci.lib.envinject.EnvInjectLogger;
+import org.jenkinsci.plugins.envinject.model.EnvInjectJobPropertyContributor;
+import org.jenkinsci.plugins.envinject.model.EnvInjectJobPropertyContributorDescriptor;
+import org.jenkinsci.plugins.envinject.service.EnvInjectActionSetter;
+import org.jenkinsci.plugins.envinject.service.EnvInjectContributorManagement;
+import org.jenkinsci.plugins.envinject.service.EnvInjectEnvVars;
+import org.jenkinsci.plugins.envinject.service.EnvInjectVariableGetter;
+import org.kohsuke.stapler.StaplerRequest;
+
+/**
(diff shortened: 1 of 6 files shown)


oleg-nenashev:
I'd like to get this feature merged. BTW, a combination of EnvInject build step and https://wiki.jenkins-ci.org/display/JENKINS/pre-scm-buildstep seems to provide this feature.

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPrebuildJobProperty.java
```diff
@@ -0,0 +1,201 @@
...
+ */
+public class EnvInjectPrebuildJobProperty<T extends Job<?, ?>> extends JobProperty<T>
+{
+
+    private EnvInjectJobPropertyInfo info = new EnvInjectJobPropertyInfo();
+    private boolean on;
+
+    @SuppressWarnings("unused")
+    public EnvInjectJobPropertyInfo getInfo()
+    {
```
oleg-nenashev: It would be great if you follow the plugin's codestyle. Brackets are being placed at the same line as the declaration in the plugin

arcivanov:
@oleg-nenashev
> BTW, a combination of EnvInject build step and https://wiki.jenkins-ci.org/display/JENKINS/pre-scm-buildstep

Actually I don't think it does, as specific stated purpose is (emphasis added):

> Creates the "prebuild" phase JobProperty section, allowing scripts __to be executed after SCM is completed but before AbstractBuildExecution.doRun kicks in__.

This allows to get a code tree sucked in and then configure things like JDK, Maven or any other build environment component based on information retrieved from the SCM.

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPrebuildJobProperty.java
arcivanov: Thanks, I'll work on the style, standby.

oleg-nenashev:
It would be great to have a 1 or 2 unit tests for the feature (just to be sure it works correctly when we apply new changes). It should not be a big deal, because it's possible to copy some bits from https://github.com/jenkinsci/envinject-plugin/pull/58. Could you add such tests?

arcivanov:
I'll add the tests

arcivanov:
@oleg-nenashev Tests added, latest functionality synced up. Happy? :smile:

oleg-nenashev:
@arcivanov 
Thanks a lot for your efforts. The PR looks good to me. Somebody may be concerned by another checkbox in the job property, but IMO it's OK in the current case (and can be a refactoring subject for envinject-2.0). 

Adding the PR to the next major release milestone. It would be great if you squash commits

arcivanov:
@oleg-nenashev squashed, commit message updated to reflect fixed status

arcivanov:
@oleg-nenashev bump

oleg-nenashev:
@daniel-beck 
> Isn't it time the UI was changed to something more like Git Plugin's Additional Behaviors, or how Email-ext changed the per-trigger recipients? It's getting ridiculous with the boxes.

It definitely makes sense (as well as the rework of the internal layout). BTW it should be done as a general envinject-2.0 effort.

I'm pretty OK to merge the PR now, but I would plan to rework it during the 2.0 rework and probably make the incompatible change.

arcivanov:
I agree that UI rework is out of scope for this PR - it was my intention to only introduce additional functionality.

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPrebuildJobProperty.java
```diff
@@ -0,0 +1,390 @@
...
+            for (EnvInjectJobPropertyContributor contributor : contributors) {
+                contributionVariables.putAll(contributor.getEnvVars(build, listener));
+            }
+        }
+        return contributionVariables;
+    }
+
+    
+    @Deprecated
+    private transient boolean injectGlobalPasswords;
```
daniel-beck: @dummy-lanwen-bot (and @lanwen ) This should `code format` the `@deprecated` otherwise the person behind that stupid user name will hate you.

lanwen: @daniel-beck I've turned on this like an experiment, sorry for spam. (Already turned off). 

I'll create very light profile for jenkins plugins to avoid spam in future, as of it too many violations of code style in any place :(

PR #37, opened 2015-01-24
[WIP][JENKINS-26583] Env-inject suppress variables contributed by extension points

olivergondza:
[JENKINS-26583](https://issues.jenkins-ci.org/browse/JENKINS-26583)

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/envinject/EnvInjectActionTest.java
+++ b/src/test/java/org/jenkinsci/plugins/envinject/EnvInjectActionTest.java
@@ -41,7 +41,6 @@ import hudson.tasks.Shell;
 
 import java.io.IOException;
 import java.util.Map;
-import java.util.concurrent.ExecutionException;
 
 import org.junit.Rule;
 import org.junit.Test;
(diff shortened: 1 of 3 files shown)


src/test/java/org/jenkinsci/plugins/envinject/EnvInjectActionTest.java
```diff
@@ -0,0 +1,199 @@
...
+public class EnvInjectActionTest {
+
+    @Rule public JenkinsRule j = new JenkinsRule();
+
+    @Test public void doNotOverrideWrapperEnvVar() throws Exception {
+        FreeStyleProject p = setupProjectWithDefaultEnvValue();
+
+        p.getBuildWrappersList().add(new ContributingWrapper("DISPLAY", "BUILD_VAL"));
+
+        validate(p);
```
lanwen: please don't hide validation process in "mistic guest" this is antipattern of test writing. Use hamcrest matchers!

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectActionTest.java
```diff
@@ -0,0 +1,199 @@
...
+import org.jvnet.hudson.test.CaptureEnvironmentBuilder;
+import org.jvnet.hudson.test.JenkinsRule;
+import org.jvnet.hudson.test.TestBuilder;
+import org.jvnet.hudson.test.TestExtension;
+
+public class EnvInjectActionTest {
+
+    @Rule public JenkinsRule j = new JenkinsRule();
+
+    @Test public void doNotOverrideWrapperEnvVar() throws Exception {
```
lanwen: In other places annotations are on top of methods - this increases of readability of code
```java
@Test
public void test...()
```
(same with rules)

olivergondza:
@lanwen, I have addressed your concerns. Using hamcrest for validation* would require bumping the inherited version.

\*) there are no `Matcher#describeMismatch`( or `TypesafeMatcher#describeMismatchSafely`) so it would actually made defect localization weaker.

lanwen:
@olivergondza thanks for quick reply!
Think junit version can be bumped to 4.12 without any doubt. This version is stable and used in production testing half a year. Test should use all features of new version of this lib and hamcrest lib

oleg-nenashev:
I would be also happy to see it merged

olivergondza:
The fix does not work is it should, I did not managed to untangle the injection logic IIRC.

jukkasi:
+1

oleg-nenashev:
Deleting garbage comments from @dummy-lanwen-bot

oleg-nenashev:
+1 for integrating it into EnvInject-2

DanielRedOak:
I think we're seeing this problem in combination with the rvm plugin.  GEM_PATH ends up with pre-rvm init values so it doesnt work.  The job isnt using any envinject features, but it is still causing issues.

oleg-nenashev:
I will check if I can integrate it towards the 2.1 release

fiznool:
The 2.1 release came and went without this getting merged... any plans on getting this in to a future release? It's blocking my use of the nodejs plugin 😞

oleg-nenashev:
@fiznool There is a plan for it, but it is not trivial to fix it without breaking changes in other usa-cases. Right now I do not have so much spare time to work on EnvInject (the plugin is waiting for adoption), and effectively this change priority is being defined by the internal priorities of the company I work for. These priorities shift sometimes, hence I cannot commit on the ETA.

If somebody proposes a fix, I will be happy to review it.

fiznool:
Thanks for the update. I know how hard it is to maintain OSS. Hope you find
some time to carry out the necessary work at some point soon!

oleg-nenashev:
Created https://github.com/jenkinsci/envinject-plugin/pull/124, which really fixes some bits

oleg-nenashev:
Closing in favor of #125, which offers a partially working fix covering JENKINS-26583.
2 tests created by @olivergondza still fails, but I think they can be fixed later since the fix logic is quite different (see #124).

PR #40, opened 2015-03-09
[JENKINS-23447] related , avoid NPE if there are no getSensibleVariables()

dshvedchenko:


src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginAction.java
```diff
@@ -22,7 +22,7 @@ public Object getTarget() {
         return new EnvInjectVarList(Maps.transformEntries(envMap,
                 new Maps.EntryTransformer<String, String, String>() {
                     public String transformEntry(String key, String value) {
-                        return getSensibleVariables().contains(key) ? "********" : value;
+                        return getSensibleVariables() != null && getSensibleVariables().contains(key) ? "********" : value;
```
oleg-nenashev: In order to be 100%-sure about the value, call the method and assign its output to a variable before checking return values

dshvedchenko:
@oleg-nenashev , is there any chances that plugin 1.92 will be released soon ?

oleg-nenashev:
It makes sense to release a new version ASAP due to its possible impact.
I'll try to perform some additional testing tomorrow and then release the new version

gboissinot:
Thank you for your contribution. Feel free to release a new version.

oleg-nenashev:
The new version has been released.
BTW, the bug seems to be quite painful, because the code in envinject-lib is not protected from the null list. Currently I'm annotating the code in order to find the real cause of the issue

dshvedchenko:
thank you for releasing it

oleg-nenashev:
@dshvedchenko 
Could you list all Project type plugins you use?
I suppose the issue happens for a plugin, which overrides <code>AbstractBuild::getSensitiveBuildVariables()</code> and returns <code>null</code>. The core's Javadoc does not describe such situation

dshvedchenko:
those are option when for New Item: hope it will help
* Freestyle project
* MultiJob Project
* Maven project
* External Job
* Folder
* Multi-configuration project

oleg-nenashev:
OK, thanks a lot. I'll submit patches to both Jenkins core and envinject-lib

PR #45, opened 2015-04-14
[FIXED JENKINS-28188] - Fix for plugin compatibility when used anonymous class cause

lanwen:
Fixes [JENKINS-28188](https://issues.jenkins-ci.org/browse/JENKINS-28188)
Example for reproduce (test added):
```java
 FreeStyleBuild build = jenkins.createFreeStyleProject().scheduleBuild2(0, new Cause() {
            @Override
            public String getShortDescription() {
                return "This build was started by a hobbit Bilbo. Bilbo Baggins";
            }
        }).get();
```

Exception:
```
[EnvInject] - Loading node environment variables.
FATAL: String index out of range: -1
http://stacktrace.jenkins-ci.org/search?query=java.lang.StringIndexOutOfBoundsException: String index out of range: -1
at java.lang.String.substring(String.java:1875) <http://stacktrace.jenkins-ci.org/search/?query=java.lang.String.substring&entity=method>
at org.jenkinsci.plugins.envinject.service.BuildCauseRetriever.buildCauseEnvironmentVariables(BuildCauseRetriever.java:71) <http://stacktrace.jenkins-ci.org/search/?query=org.jenkinsci.plugins.envinject.service.BuildCauseRetriever.buildCauseEnvironmentVariables&entity=method>
at org.jenkinsci.plugins.envinject.service.BuildCauseRetriever.getTriggeredCause(BuildCauseRetriever.java:40) <http://stacktrace.jenkins-ci.org/search/?query=org.jenkinsci.plugins.envinject.service.BuildCauseRetriever.getTriggeredCause&entity=method>
at org.jenkinsci.plugins.envinject.service.EnvInjectVariableGetter.getBuildVariables(EnvInjectVariableGetter.java:111) <http://stacktrace.jenkins-ci.org/search/?query=org.jenkinsci.plugins.envinject.service.EnvInjectVariableGetter.getBuildVariables&entity=method>
at org.jenkinsci.plugins.envinject.EnvInjectListener.setUpEnvironmentWithoutJobPropertyObject(EnvInjectListener.java:231) <http://stacktrace.jenkins-ci.org/search/?query=org.jenkinsci.plugins.envinject.EnvInjectListener.setUpEnvironmentWithoutJobPropertyObject&entity=method>
[...]

PR #46, opened 2015-04-22
[JENKINS-20334] - envInject leaves in backslash for #, !, =, : 

rafaelrezend:
The main motivation for this fix was the issue [JENKINS-20334](https://issues.jenkins-ci.org/browse/JENKINS-20334), although it could solve other issues which are unknown to me at the moment.

Since **EnvInject** plugin is expected to read from a *.properties* file, it should use the very same convention as defined in *java.util.Properties*:
> *[...] Characters that cannot be directly represented in this encoding can be written using Unicode escapes as defined in [section 3.3 of The Java™ Language Specification](https://docs.oracle.com/javase/specs/jls/se7/html/jls-3.html#jls-3.3) [...]*

The *load()* and *store()* methods provided by the *Properties* class natively handles these twists during serialization/deserialization. So, why not use it directly instead of *hudson.Util.loadFile(…)* followed by manual “unescaping” of the properties (using *processPath(…)*)?
At this point, the local/remote nature of the file has already been resolved.

The implementation in this Pull Request was tested with several properties and special characters. It was also tested in a master/slave configuration (slaves on virtual machine).

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/envinject/service/PropertiesLoader.java
+++ b/src/main/java/org/jenkinsci/plugins/envinject/service/PropertiesLoader.java
@@ -31,32 +31,6 @@ public class PropertiesLoader implements Serializable {
      * @throws EnvInjectException
      * @throws IOException 
      */
-//    public Map<String, String> getVarsFromPropertiesFile(File propertiesFile, Map<String, String> currentEnvVars) throws EnvInjectException {
-//
-//        if (propertiesFile == null) {
-//            throw new NullPointerException("The properties file object must be set.");
-//        }
-//        if (!propertiesFile.exists()) {
-//            throw new IllegalArgumentException("The properties file object must be exist.");
-//        }
-//
-//        Map<String, String> result = new LinkedHashMap<String, String>();
-//
-//        SortedProperties properties = new SortedProperties();
-//        try {
-//            String fileContent = Util.loadFile(propertiesFile);
-//            String fileContentResolved = Util.replaceMacro(fileContent, currentEnvVars);
-//            fileContentResolved = processPath(fileContentResolved);
-//            properties.load(new StringReader(fileContentResolved));
-//        } catch (IOException ioe) {
-//            throw new EnvInjectException("Problem occurs on loading content", ioe);
-//        }
-//        for (Map.Entry<Object, Object> entry : properties.entrySet()) {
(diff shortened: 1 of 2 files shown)


rafaelrezend:
> *Error Message*
> expected:</[tmp/hudson2388473214719746968test/customWorkspace]/materialize_workspa...> but was:</[scratch/jenkins/workspace/plugins/envinject-plugin]/materialize_workspa...>

The issue was also experienced in previous Pull Requests. It does not happen locally.
[lanwen](https://github.com/lanwen) seems to have a fix for it in commit [8141652](https://github.com/lanwen/envinject-plugin/commit/81416520412995432ea9fe444c635e578dd3aa8d) of the pending Pull Request #45.

lanwen:
@rafaelrezend Problem with test is that host jenkins overrides `workspace` env in test-jenkins. So i've assume no any `GIT_PR_FROM` env vars is set. I couldn't find better solution

oleg-nenashev:
The change is reasonable, but it will cause massive regressions due to the change of the processing order. Example: it will be impossible to specify multiple env vars in a single external variable. E.g. <code>${FILE_CONTENTS}</code> becomes an invalid input entry.

rafaelrezend:
@oleg-nenashev Sorry, I think I didn't get the use case. I've added variables to both key and values in the test cases, but I didn't understand yours. More info?

oleg-nenashev:
@rafaelrezend 
Sorry for a delay. I've mentioned the following use-case:

envinject_1_var

It works well on current installations. I've definitely used such solution several times on previous installations.

rafaelrezend:
Closing rejected Pull Request. Reason: see comments from @oleg-nenashev .

oleg-nenashev:
@rafaelrezend Thanks for your efforts. Just to clarify my position... I think this improvement may be useful.. Even if the PR cannot be merged in the current state, it can be improved somehow (advanced options, etc.), so feel free to propose another version if you still want to get the issue fixed. If no, probably somebody starts working on the issue later

rafaelrezend:
@oleg-nenashev I might come back with a solution, but not so soon :-)
I did not predict the use case you mentioned (btw, this should be added to the automated tests in the future), so, thanks for the help. For while I'm using a workaround to avoid the issue [JENKINS-20334](https://issues.jenkins-ci.org/browse/JENKINS-20334): I'm writing URLs into the properties file directly, without using Java Properties. So, no escape is applied there, and it works for me right now :-)

PR #49, opened 2015-05-02
JENKINS-28188 StringIndexOutOfBoundsException when no root cause

dcoraboeuf:
When a build has no root cause (possible when firing the job programmatically), the code at `org.jenkinsci.plugins.envinject.service.BuildCauseRetriever#buildCauseEnvironmentVariables` produces a `StringIndexOutOfBoundsException` because the substring method is applied to an empty string.

The fix I propose is very basic and could be simplified using Guava - but I didn't date to use it in this plugin.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/envinject/service/BuildCauseRetriever.java
+++ b/src/main/java/org/jenkinsci/plugins/envinject/service/BuildCauseRetriever.java
@@ -59,20 +59,15 @@ public class BuildCauseRetriever {
 
     private static Map<String, String> buildCauseEnvironmentVariables(String envBase, Collection<String> causeNames) {
         Map<String, String> triggerVars = new HashMap<String, String>();
-        StringBuilder all = new StringBuilder();
-        int count = 0;
+        List<String> all = new ArrayList<String>();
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
I suppose <code>StringUtils.join()</code> could resolve the issue in a more elegant way. BTW, we would still need a cycle (and null/empty string checks), so LGTM

dcoraboeuf:
Yes, StringUtils, of course :( I'll adjust accordingly.

dcoraboeuf:
Done.

dcoraboeuf:
Tests are failing in Cloudbees, not locally. I'll have a look later on.

dcoraboeuf:
The failure has nothing to do with this PR - it was already there before.

oleg-nenashev:
LGTM. Thanks for the fix

lanwen:
This is same as #45

oleg-nenashev:
Closing in a favor of #45 
@dcoraboeuf thanks for you efforts. The original title in #45 was too narrow specific, so unfortunately I have not noticed a duplication on early stages.

lanwen:
fixed test in #51

PR #51, opened 2015-05-15
fix for failed build on cloudbees (or other CI) 

lanwen:
When `WORKSPACE` env var already set, it replaces the custom build property injection.

As of you don't usually run jenkins inside of jenkins, this test can simply remove this var from env vars.

Also: simplify test-code without change of logic

Summon @oleg-nenashev for review

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/envinject/EnvInjectBuildWrapperTest.java
+++ b/src/test/java/org/jenkinsci/plugins/envinject/EnvInjectBuildWrapperTest.java
@@ -1,27 +1,31 @@
 package org.jenkinsci.plugins.envinject;
 
-import static org.junit.Assert.assertEquals;
-import hudson.Util;
+import hudson.EnvVars;
 import hudson.model.FreeStyleBuild;
 import hudson.model.FreeStyleProject;
-import hudson.model.Hudson;
-import hudson.model.Result;
-import junit.framework.Assert;
-
-import org.jenkinsci.lib.envinject.EnvInjectAction;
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/plugins/envinject/EnvInjectBuildWrapperTest.java
```diff
@@ -61,47 +65,35 @@ public void injectFromScript() throws Exception {
...
-        Assert.assertNotNull(envVars);
-        String resolvedValue = envVars.get(customEnvVarName);
-        Assert.assertNotNull(resolvedValue);
-
-        //3-- Test equals
-        Assert.assertEquals(expectedCustomEnvVarValue, resolvedValue);
+        String expectedCustomEnvVarValue = replaceMacro(customEnvVarValue,
+                of("WORKSPACE", buildWorkspaceValue));
+
+        assertThat("2. Property should be resolved with custom ws in build", build,
```
oleg-nenashev: Minor: numbering of assert texts is quite uncommon

oleg-nenashev:
LGTM in general

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectBuildWrapperTest.java
lanwen: Think this testcase is quite uncommon. And should be rewritten in future. So numbering is for quick understandig the place of error in this case.

PR #53, opened 2015-07-28
[FIXED JENKINS-29667] Text description in plugin manager shouldn't be in bold

andham:
Signed-off-by: Anders Hammar <anders@hammar.net>


oleg-nenashev:
I'm not sure how it leads to the bold text, but there is no need in such tags.
:+1:

lanwen:
:+1:

PR #56, opened 2015-08-05
[JENKINS-29817] Inject environment variables for a Job

recena:
https://issues.jenkins-ci.org/browse/JENKINS-29817

- [x] Implement `EnvInjectEnvVarsContributor`
- [x] Add tests
- [x] Update the documentation (help information) for the **Properties Content** field

@reviewbybees


src/test/java/org/jenkinsci/plugins/envinject/EnvInjectEnvVarsJobTest.java
```diff
@@ -0,0 +1,33 @@
...
+        EnvInjectJobProperty envInjectJobProperty = new EnvInjectJobProperty();
+        envInjectJobProperty.setOn(true);
+        envInjectJobProperty.setInfo(jobPropertyInfo);
+        project.addProperty(envInjectJobProperty);
+
+        TaskListener listener = jenkins.createTaskListener();
+        EnvVars environment = project.getEnvironment(jenkins.getInstance(), listener);
+
+        assertNotNull(environment.get("REPO"));
+        jenkins.assertBuildStatusSuccess(project.scheduleBuild2(0).get());
```
jglick: Running the build does not really accomplish anything. You could delete this line, or make the test demonstrate that the variable is also accessible to build steps. Which reminds me that it is likely that other code in this plugin is interpreting `EnvInjectJobProperty` on a `Run` (or `AbstractBuild`) that could probably now be deleted, since `Run` environments inherit from `Job` environments.

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectEnvVarsJobTest.java
```diff
@@ -0,0 +1,33 @@
...
+import hudson.EnvVars;
+import hudson.model.FreeStyleProject;
+import hudson.model.TaskListener;
+import org.junit.Rule;
+import org.junit.Test;
+import org.jvnet.hudson.test.JenkinsRule;
+
+import static org.junit.Assert.assertNotNull;
+
+public class EnvInjectEnvVarsJobTest {
```
jglick: Arguably should be called `EnvInjectEnvVarsContributorTest`.

recena: @jglick I agree

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectEnvVarsJobTest.java
recena: @jglick Indeed. I'm going to review the test.

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectEnvVarsContributor.java
```diff
@@ -0,0 +1,27 @@
...
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
+@Extension
+public class EnvInjectEnvVarsContributor extends EnvironmentContributor {
+
+    @Override
+    public void buildEnvironmentFor(Job job, EnvVars env, TaskListener listener) throws IOException, InterruptedException {
+        for (Object property : job.getProperties().values()) {
```
oleg-nenashev: It's better to get properties by class without such cycles

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectEnvVarsContributor.java
```diff
@@ -0,0 +1,27 @@
...
+            if (property instanceof EnvInjectJobProperty) {
+                EnvInjectJobProperty jobProperty = (EnvInjectJobProperty) property;
+                EnvInjectJobPropertyInfo jobPropertyInfo = jobProperty.getInfo();
+                Map<String, String> result = jobPropertyInfo.getPropertiesContentMap(new HashMap<String, String>());
+                if (result != null) {
+                    env.putAll(result);
```
oleg-nenashev: I doubt it will work as expected if a property depends on the value of another property. A test is required to demonstrate it

oleg-nenashev:
I general I agree that such contributor mat be useful. On the other hand I cannot elaborate an impact of the change, because it may break the variable resolution order on existing installations. It may also cause a total breakage of existing jobs if they use Groovy scripts to inject variables and use build variables.

More tests and investigations are required before this PR becomes mergeable. I'm putting a :bug: on it now.

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectEnvVarsContributor.java
recena: @oleg-nenashev There is a test.

recena:
@oleg-nenashev What tests do you propose to verify that improvement (very needed) does not breaks anything? 

This PR has been tested with Subversion Plugin (master) in order to verify this kind of use case [JENKINS-29340](https://issues.jenkins-ci.org/browse/JENKINS-29340) and works perfectly.

Basically, we need the environment variables defined by EnvInject availables in `Job.getEnvironment()`.

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectEnvVarsContributor.java
recena: @oleg-nenashev I'll take a look.

oleg-nenashev:
@recena 
I'll try to respond on this evening.

oleg-nenashev:
@recena 
After digging into the code...
<code>EnvInjectInfo: getPropertiesContentMap()</code> works with property contents only (text area field). See https://github.com/jenkinsci/envinject-plugin/blob/master/src/main/java/org/jenkinsci/plugins/envinject/EnvInjectInfo.java#L42-L65. The current implementation ignores environment variables from property files, Groovy scripts and from Master node. 

Such implementation may address a narrow case from JENKINS-29817, but it is incomplete. I would not vote for merging the fix in the current state, because it may confuse the plugin users => serious :beetle: . You should implement a new method, which would generate variables without a build context.

I'm also aware about the variables resolution order. <code> buildEnvironmentFor</code> for <code>Job</code>s is being called before the resolution for runs. According to the brief code analysis, seems that the use-case with build parameter overrides will be broken by the change. The behavior is configurable via the ''Override build parameters" option. According to the manual testing, <code>AbstractBuild:getEnvironment()</code> returns wrong values after the change. It's a :bug: IMO

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectEnvVarsJobTest.java
recena: @jglick Done

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectEnvVarsJobTest.java
recena: @jglick Done

recena:
@oleg-nenashev I'm going to implement the different way to provide variables (property files, groovy scripts, etc..). I don't understand why you say that this new extension breaks anything. The rest of tests are working fine. Maybe we should review these test, Isn't that so?

recena:
@oleg-nenashev There are a lot of users using Subversion Plugin and EnvInject Plugin. We need the environment variables (injected by this plugin) availables, p.e, in polling process. What would it your proposal?

oleg-nenashev:
> I don't understand why you say that this new extension breaks anything

I just see a wrong output in the API method. This result needs more investigation. buildEnvironmentForJob() is being called before the Environment contributing actions, so probably the issue in overrides (or I'm doing something wrong in the debugger). I'll write a unit test and submit it.

>  I'm going to implement the different way to provide variables (property files, groovy scripts, etc..). 

Please be aware about Groovy ones. I suspect that many Users rely on build runtime variables there. The plugin's code correctly handles null build references, but we should ensure that the newly introduced method is not failing horribly in the case of the misusage.

oleg-nenashev:
> The rest of tests are working fine. Maybe we should review these test, Isn't that so?

Unfortunately the plugin does not contain tests for <code>EnvInjectJobProperty</code> at all :(

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectEnvVarsContributor.java
recena: @oleg-nenashev I've done some changes related to your comment. Could you review it?

recena: @oleg-nenashev I've reverted the changes because the tests were failing. I was trying this:

```
@Override
public void buildEnvironmentFor(Job job, EnvVars env, TaskListener listener) throws IOException, InterruptedException {
    EnvInjectJobProperty jobProperty = (EnvInjectJobProperty) job.getProperty(EnvInjectJobProperty.class)
    EnvInjectJobPropertyInfo jobPropertyInfo = jobProperty.getInfo();

    // Process "Properties Content"
    Map<String, String> result = jobPropertyInfo.getPropertiesContentMap(new HashMap<String, String>());
    if (result != null) {
        env.putAll(result);
    }
}
```

src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectJobProperty/help-propertiesContent.html
```diff
@@ -1,7 +1,8 @@
...
-        Give a set of key/value (KEY=VALUE).<br/>
-        All the properties name will be accessible as environment variables by their names.<br/>
-        You can use or override the properties specified in the above properties file.<br/>
-    </p>
+    <p>Give a set of key/value (one variable per line): <code>KEY=VALUE</code>.</p>
+
+    <p>All the properties name will be accessible as environment variables by their names. You can use or override the
+        properties specified in the above properties file.</p>
+
+    <p>These environment variables (<strong>only these</strong>) will be available in the context Job.</p>
```
recena: @oleg-nenashev What do you think?

oleg-nenashev: What is a <code>context Job</code>? I don't know such term. In any case it makes sense to add several sentences, which would allow users to understand what's going on here.

src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectJobProperty/help-propertiesContent.html
```diff
@@ -1,7 +1,12 @@
...
+
+    <p>All the properties name will be accessible as environment variables by their names. You can use or override the
+        properties specified in the above properties file.</p>
+
+    <p>These environment variables (<strong>only these</strong>) will be available in the Job. Use cases:</p>
+    <ul>
+        <li>Validate Subversion Repository with a URL like this: <code>http://server/svn/$REPO</code></li>
+        <li>Subversion SCM Polling</li>
+    </ul>
 </div>
```
recena: @oleg-nenashev Better?

oleg-nenashev: My proposals:
```
... will be available in the Job environment independently from the build. There are following use-cases: SCM polling, validation of fields depending on environment variables (e.g. http://server/svn/$REPO), custom workspace management, etc. 

Limitations: The field may contain variables, which are not resolvable outside the Build context (e.g. BUILD_NUMBER, WORKSPACE, etc.), the plugin [injects empty value/ignores the field/or something else]). This behavior should be taken into account by job editors.
```

Please also add a proper test-case for the limitation

recena: @oleg-nenashev Thanks so much for your help. I'm going to investigate the **limitations**.

recena: @oleg-nenashev I don't understand what you mean with **limitations**. Could you show an example the variable defined in this field *not resolvable*? Do you mean something like this `VAR1=WORKSPACE`?

oleg-nenashev: @recena 
Yes, WORKSPACE var is a valid example for this limitation

recena:
@oleg-nenashev What do you thing?

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectEnvVarsContributor.java
oleg-nenashev: Null check is missing. If a job has no <code> EnvInjectJobProperty</code>, your test will fail with NPE

src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectBuildWrapper/config.jelly
```diff
@@ -35,6 +21,20 @@
                 value="${instance.info.groovyScriptContent}"/>
     </f:entry>
 
+    <f:entry field="propertiesFilePath"
```
oleg-nenashev: Why did you move this section? How is it related to this PR?

src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectBuildWrapper/help-propertiesContent.html
```diff
@@ -0,0 +1,6 @@
+<div>
```
oleg-nenashev: Seems you've forgotten to remove the original help files

recena: @oleg-nenashev We need both help files because we can use this plugin in different ways:

* Prepare an environment for the run: *...set an environment for the build job* (before)
* Inject environment variables to the build process: *... to the whole build process.* (during)

Please, find attached this screenshot:

jenkins-29817

oleg-nenashev: Now I get the reason. In such case you should reference a text with extra details by URL. It will prevent confusions of users

src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectBuildWrapper/config.jelly
recena: @oleg-nenashev Indeed this change is not directly related to PR but I considered it would be helpful to show the form fields in the same order in both sections. Don't make sense to use a different order.

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectEnvVarsContributor.java
recena: @oleg-nenashev AFAIK, if you are using the plugin (checkbox is clicked), there is always a `EnvInjectJobProperty`. And I've verified that `job.getProperties()` can never be `null`.

oleg-nenashev: job.getProperty() may return null if you run the method agains the loaded job, which has been saved before the EnvInject installation. 

<code>job.getProperties()</code> is never null, but IMO you're using a wrong pattern to lookup for properties. It may become performance-innefective in future Jenkins versions

src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectBuildWrapper/help-propertiesContent.html
recena: @oleg-nenashev I don't know what you mean with *"...reference a text with extra details by URL."*

oleg-nenashev: I just mean it makes sense to put the extended help message to resources and reference it explicitly

src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectBuildWrapper/help-propertiesContent_ja.html
```diff
@@ -0,0 +1,7 @@
+<div>
+    <p>
+        KEY=VALUE形式で、キーと値を設定します。<br />
+        環境変数にプロパティの名前でアクセスできます。<br />
+        上記のプロパティファイルで設定したプロパティを使用したり、上書きすることも可能です。
```
oleg-nenashev: I suspect it's the original message. It makes sense to remove it (users will get the English text) or ask somebiody like @kohsuke to translate it. Otherwise users with Japanese locale will get a message without sensitive details

src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectBuildWrapper/config.jelly
oleg-nenashev: NIT: "Properties content" is the most widely used section, so I would use another direction. BTW, the interface should be reworked to show only the required section (extension point?). Such rework is not a part of the PR, so I'd prefer to submit it as another PR

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectEnvVarsContributor.java
recena: @oleg-nenashev I don't know how job.getProperty() can be `null` if this method is invoked only if `job.getProperties().values()` returns values.

Do you prefer to use `job.getProperty(EnvInjectJobProperty.class)`?

oleg-nenashev: Yes. And it can return the null value ;)

pom.xml
```diff
@@ -5,7 +5,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>1.532</version>
+        <version>1.532.3</version>
```
oleg-nenashev: Yes, the original version makes me sad as well (unit tests do not work reliably)

src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectBuildWrapper/config.jelly
recena: @oleg-nenashev We can discuss about it out of the PR.

src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectBuildWrapper/help-propertiesContent.html
recena: @olamy In this case, I prefer to leave at that.

oleg-nenashev:
:bug: due to the unrelated change in <code>src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectBuildWrapper/config.jelly</code>, which is not related to this PR. It should be decoupled to a separate pull-request

src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectBuildWrapper/config.jelly
oleg-nenashev: As a summary of the discussion, I still think it should be moved to another PR:
1) It is completely unrelated to the PR
2) The change impacts the usability and UX IMO, because it changes the original layout

amuniz:
@oleg-nenashev Is that really a bug? Is preventing to use the plugin somehow? Since it's not really a functional change and does not affect in any way to the behavior being modified in this PR, it's ok for me. Moreover it's a minor change that you can quick see that is only a code block moved some lines below.

src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectBuildWrapper/help-propertiesContent.html
oleg-nenashev: OK, agreed

oleg-nenashev:
Please also address a comment regarding the Japanese help

src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectBuildWrapper/help-propertiesContent_ja.html
lanwen: maybe just use some [translate service](https://translate.yandex.com/)?

oleg-nenashev:
:bee: 
Could you please squash commits? After that we will be able to merge it according to the policy. I'm planning to release 0.92-beta1 on the weekend (or today if everything goes right)

recena:
@oleg-nenashev done

recena:
@oleg-nenashev By the way, Is there an offical maintainer? /cc @gboissinot

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectEnvVarsContributorTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+        EnvInjectJobPropertyInfo jobPropertyInfo = new EnvInjectJobPropertyInfo(null, "VAR1=${WORKSPACE}\nVAR2=${JOB_NAME}", null, null, null, false);
+        EnvInjectJobProperty envInjectJobProperty = new EnvInjectJobProperty();
+        envInjectJobProperty.setOn(true);
+        envInjectJobProperty.setInfo(jobPropertyInfo);
+        project.addProperty(envInjectJobProperty);
+
+        TaskListener listener = jenkins.createTaskListener();
+        EnvVars environment = project.getEnvironment(jenkins.getInstance(), listener);
+        assertEquals("${WORKSPACE}", environment.get("VAR1"));
```
oleg-nenashev: This check is failing on my local laptop. The result is null

oleg-nenashev: fixed by clean build, please ignore

PR #57, opened 2015-08-09
[JENKINS-29867] - Permissions engine + global option for disabling the Injected vars

oleg-nenashev:
https://issues.jenkins-ci.org/browse/JENKINS-29867

In several cases Jenkins users may expose passwords as a plain text in "Injected variables" action (e.g. password reassigning for a non-sensitive variable). It may make sense to add an option, which completely prohibits the exposure of environment variables.

- [x] - ViewVars permissions
- [x] - Global option, which allows to hide envinject actions
- [x] - Unit tests
- [x] - Bump the dependency on https://github.com/jenkinsci/envinject-lib after the release of https://github.com/jenkinsci/envinject-lib/pull/6

Depends on https://github.com/jenkinsci/envinject-lib/pull/6

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -38,7 +38,7 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <maven.compiler.source>1.6</maven.compiler.source>
         <maven.compiler.target>1.6</maven.compiler.target>
-        <envinject.lib.version>1.22</envinject.lib.version>
+        <envinject.lib.version>1.23</envinject.lib.version>
         <ivy.plugin.version>1.21</ivy.plugin.version>
         <junit.version>4.12</junit.version>
         <mockito.version>1.10.19</mockito.version>
--- a/src/main/java/org/jenkinsci/plugins/envinject/EnvInjectAction.java
+++ b/src/main/java/org/jenkinsci/plugins/envinject/EnvInjectAction.java
@@ -15,6 +15,7 @@ import java.util.Set;
 
 /**
  * @author Gregory Boissinot
+ * @deprecated Replaced by {@link EnvInjectPluginAction}.
  */
 @Deprecated
 public class EnvInjectAction implements Action, StaplerProxy {
@@ -45,20 +46,38 @@ public class EnvInjectAction implements Action, StaplerProxy {
         return UnmodifiableMap.decorate(envMap);
     }
 
+    @Override
     public String getIconFileName() {
+        if (!EnvInjectPlugin.canViewInjectedVars(build)) {
+            return null;
+        }
         return "document-properties.gif";
     }
 
+    @Override
     public String getDisplayName() {
         return "Environment Variables";
     }
 
+    @Override
     public String getUrlName() {
+        if (!EnvInjectPlugin.canViewInjectedVars(build)) {
+            return null;
+        }
         return URL_NAME;
     }
 
+    protected AbstractBuild getBuild() {
+        return build;
+    }
+    
+    @Override
     public Object getTarget() {
         final Set sensitiveVariables = build.getSensitiveBuildVariables();
+        if (!EnvInjectPlugin.canViewInjectedVars(build)) {
+            return EnvInjectVarList.HIDDEN;
+        }
+        
         return new EnvInjectVarList(Maps.transformEntries(envMap,
                 new Maps.EntryTransformer<String, String, String>() {
                     public String transformEntry(String key, String value) {
--- a/src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPlugin.java
+++ b/src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPlugin.java
@@ -1,0 +1,113 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.jenkinsci.plugins.envinject;
+
+import hudson.Plugin;
+import hudson.model.Run;
+import hudson.security.Permission;
+import hudson.security.PermissionGroup;
+import hudson.security.PermissionScope;
+import java.util.logging.Logger;
+import javax.annotation.Nonnull;
+import jenkins.model.Jenkins;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+import org.kohsuke.stapler.export.ExportedBean;
+
+/**
+ * Stores permissions for EnvInject plugin.
+ * @author Oleg Nenashev
+ * @since 1.92
+ */
+@ExportedBean
+public class EnvInjectPlugin extends Plugin {
+
+    private final static Logger LOGGER = Logger.getLogger(EnvInjectPlugin.class.getName());
+
+    public static final PermissionGroup PERMISSIONS = new PermissionGroup(EnvInjectPlugin.class, Messages._envinject_permissions_title());
+
+    /**
+     * Allows to view injected variables.
+     * Even Jenkins admins may have no such permission in particular installations.
+     */
+    public static final Permission VIEW_INJECTED_VARS = new Permission(PERMISSIONS, "ViewVars", Messages._envinject_permissions_viewVars_description(), null, PermissionScope.RUN);
+     
+    /**
+     * Retrieves the plugin instance.
+     * @return {@link EnvInjectPlugin}
+     * @throws IllegalStateException the plugin has not been loaded yet
+     */
+    public static @Nonnull EnvInjectPlugin getInstance() {
+        Jenkins j = Jenkins.getInstance();
+        EnvInjectPlugin plugin = j != null ? j.getPlugin(EnvInjectPlugin.class) : null;
+        if (plugin == null) { // Fail horribly
+            // TODO: throw a graceful error
+            throw new IllegalStateException("Cannot get the plugin's instance. Jenkins or the plugin have not been initialized yet");
+        }
+        return plugin;
+    }
+    
+    /*package*/ void onConfigChange(@Nonnull EnvInjectPluginConfiguration config) {
+       VIEW_INJECTED_VARS.setEnabled(config.isEnablePermissions()); 
+    }
+
+    @Nonnull
+    public EnvInjectPluginConfiguration getConfiguration() {
+        final EnvInjectPluginConfiguration config = EnvInjectPluginConfiguration.getInstance();
+        return config != null ? config : EnvInjectPluginConfiguration.getDefault();
+    }
+    
+    /**
+     * Checks if the current user can view injected variables in the run.
+     * @param run Run to be checked
+     * @return true if the injected variables can be displayed.
+     */
+    @Restricted(NoExternalUse.class)
+    public static boolean canViewInjectedVars(@Nonnull Run<?,?> run) {
+        // We allow security engines to block the output
+        if (VIEW_INJECTED_VARS.getEnabled() &&  !run.hasPermission(VIEW_INJECTED_VARS)) {
+            return false;
+        }
+        
+        // Last check - global configs
+        final EnvInjectPluginConfiguration configuration = getInstance().getConfiguration();
+        return !configuration.isHideInjectedVars();
+    }
+    
+    @Override 
+    public void start() throws Exception {
+        VIEW_INJECTED_VARS.setEnabled(getConfiguration().isEnablePermissions());
+    }  
+    
+    // TODO: Replace by Jenkins::getActiveInstance() in 1.590+
+    @Nonnull
+    @Restricted(NoExternalUse.class)
+    public static Jenkins getJenkinsInstance() throws IllegalStateException {
+        final Jenkins jenkins = Jenkins.getInstance();
+        if (jenkins == null) {
+            throw new IllegalStateException("Jenkins instance is not ready");
+        }
+        return jenkins;
+    }
+}
--- a/src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
+++ b/src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
@@ -1,0 +1,116 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2015, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.jenkinsci.plugins.envinject;
+
+import com.google.common.annotations.VisibleForTesting;
+import hudson.Extension;
+import hudson.XmlFile;
+import java.io.File;
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+import jenkins.model.GlobalConfiguration;
+import jenkins.model.Jenkins;
+import net.sf.json.JSONObject;
+import static org.jenkinsci.plugins.envinject.EnvInjectPlugin.getJenkinsInstance;
+import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.StaplerRequest;
+
+/**
+ * Configuration of {@link EnvInjectPlugin}.
+ * @author Oleg Nenashev
+ * @since 1.92
+ */
+@Extension
+public class EnvInjectPluginConfiguration extends GlobalConfiguration {
+    
+    private static final EnvInjectPluginConfiguration DEFAULT = 
+            new EnvInjectPluginConfiguration(false, false);
+            
+    private boolean hideInjectedVars;
+
+    private boolean enablePermissions;
+
+    public EnvInjectPluginConfiguration() {
+        load();
+    }
+    
+    public EnvInjectPluginConfiguration(boolean hideInjectedVars, boolean enablePermissions) {
+        this.hideInjectedVars = hideInjectedVars;
+        this.enablePermissions = enablePermissions;
+    }
+
+    public boolean isHideInjectedVars() {
+        return hideInjectedVars;
+    }
+
+    public boolean isEnablePermissions() {
+        return enablePermissions;
+    }
+ 
+    /**
+     * Gets the default configuration of {@link EnvInjectPlugin}
+     * @return Default configuration
+     */
+    public static final @Nonnull EnvInjectPluginConfiguration getDefault() {
+        return DEFAULT;
+    }
+
+    @Override
+    protected XmlFile getConfigFile() {
+        return new XmlFile(Jenkins.XSTREAM, new File(getJenkinsInstance().getRootDir(), 
+                "envinject-plugin-configuration.xml"));
+    }
+
+    @Override
+    public boolean configure(StaplerRequest req, JSONObject json) throws FormException {
+        final boolean newEnablePermissions = json.getBoolean("enablePermissions");
+        final boolean newHideInjectedVars = json.getBoolean("hideInjectedVars");
+        return configure(newHideInjectedVars, newEnablePermissions);
+    }
+    
+    /**
+     * Configuration method for testing purposes.
+     * @param hideInjectedVars Hide injected variables actions
+     * @param enablePermissions Enables permissions in {@link EnvInjectPlugin}.
+     * @return true if the configuration successful
+     * @throws IllegalStateException Cannot retrieve the plugin config instance
+     */
+    @VisibleForTesting
+    /*package*/ static boolean configure(boolean hideInjectedVars, boolean enablePermissions)  {
+        EnvInjectPluginConfiguration instance = getInstance();
+        if (instance == null) {
+            throw new IllegalStateException("Cannot retrieve the plugin config instance");
+        }
+        instance.hideInjectedVars = hideInjectedVars;
+        instance.enablePermissions = enablePermissions;
+        EnvInjectPlugin.getInstance().onConfigChange(instance);
+        instance.save();
+        return true;
+    }
+    
+    @CheckForNull
+    public static EnvInjectPluginConfiguration getInstance() {
+        return EnvInjectPluginConfiguration.all().get(EnvInjectPluginConfiguration.class);
+    }
+}
--- a/src/main/java/org/jenkinsci/plugins/envinject/EnvInjectVarList.java
+++ b/src/main/java/org/jenkinsci/plugins/envinject/EnvInjectVarList.java
@@ -10,6 +10,7 @@ import javax.servlet.ServletOutputStream;
 import java.io.IOException;
 import java.io.Serializable;
 import java.io.StringWriter;
+import java.util.Collections;
 import java.util.Map;
 import java.util.TreeMap;
 
@@ -20,6 +21,12 @@ import java.util.TreeMap;
 public class EnvInjectVarList implements Serializable {
 
     private Map<String, String> envVars = new TreeMap<String, String>();
+    
+    /**
+     * Empty variables list, which should be returned if the variables are hidden
+     * due to the security settings.
+     */
+    public static final EnvInjectVarList HIDDEN = new Hidden();
 
     public EnvInjectVarList(Map<String, String> envMap) {
         if (envMap != null) {
@@ -124,6 +131,18 @@ public class EnvInjectVarList implements Serializable {
         sb.delete(0, 1);
         outputStream.write(sb.toString().getBytes());
         outputStream.write("]}}".getBytes());
+    }
+    
+    //TODO: Throw errors in responses?
+    /**
+     * Implements an {@link EnvInjectVarList}, which does not provide any variables.
+     */
+    private static class Hidden extends EnvInjectVarList {
+        private static final long serialVersionUID = 1L;
+
+        public Hidden() {
+            super(Collections.<String,String>emptyMap());
+        }  
     }
 
 }
--- a/src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration/config.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration/config.jelly
@@ -1,0 +1,37 @@
+<!--
+
+    The MIT License (MIT)
+
+    Copyright (c) 2015, CloudBees, Inc.
+
+    Permission is hereby granted, free of charge, to any person obtaining a copy
+    of this software and associated documentation files (the "Software"), to deal
+    in the Software without restriction, including without limitation the rights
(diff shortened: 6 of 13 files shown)


pom.xml
```diff
@@ -38,7 +38,7 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <maven.compiler.source>1.6</maven.compiler.source>
         <maven.compiler.target>1.6</maven.compiler.target>
-        <envinject.lib.version>1.22</envinject.lib.version>
+        <envinject.lib.version>1.23-SNAPSHOT</envinject.lib.version>
```
oleg-nenashev: Should be bumped before the release

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectPluginActionTest.java
```diff
@@ -0,0 +1,155 @@
...
+    
+    private boolean hasPermission(User user, AccessControlled item, Permission permission ) 
+            throws AssertionError {
+        SecurityContext initialContext = null;
+        try {
+            Authentication auth = user.impersonate();
+            initialContext = hudson.security.ACL.impersonate(auth);
+            return item.hasPermission(permission);
+        } catch (UsernameNotFoundException ex) {
+            throw new AssertionError("The specified user is not found", ex);
```
lanwen: use `fail(...)` instead of throwing assertion error

oleg-nenashev: fail(...) does not allow to attach the nested exception. In any case, it just wraps the throw call, so I don't see an issue with the current approach

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectAction.java
```diff
@@ -46,6 +47,9 @@ public void overrideAll(Map<String, String> all) {
     }
 
     public String getIconFileName() {
+        if (!EnvInjectPlugin.canViewInjectedVars(build)) {
+            return null;
```
amuniz: This method would require ```@CheckForNull```

oleg-nenashev: This annotation is being inherited from the <code>Action::getIconFileName()</code>, hence there is no need to refer it again. BTW the <code>@Override</code> annotation would be useful to make it more clear

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectPluginActionTest.java
lanwen: so why not just don't catch it?

oleg-nenashev: I'm not sure about it. Missing username at this stage is an impossible state meaning that something is borked by a parallel thread. I'd prefer to fail the test immediately in such case => Runtime Assertion error.

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
```diff
@@ -0,0 +1,84 @@
...
+    public boolean isHideInjectedVars() {
+        return hideInjectedVars;
+    }
+
+    public boolean isEnablePermissions() {
+        return enablePermissions;
+    }
+ 
+    /**
+     * Gets the default configuration of {@link DockerTraceabilityPlugin}
```
amuniz: DockerTraceabilityPlugin?

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectVarList.java
```diff
@@ -32,7 +39,7 @@ public EnvInjectVarList(Map<String, String> envMap) {
     public Map<String, String> getEnvMap() {
         return envVars;
     }
-
+
```
amuniz: whitespaces typo

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectPluginActionTest.java
```diff
@@ -0,0 +1,155 @@
...
+ */
+public class EnvInjectPluginActionTest {
+        
+    @Rule
+    public JenkinsRule j = new JenkinsRule();
+    
+    private FreeStyleProject p;
+    private CaptureEnvironmentBuilder capture;
+    
+    @Before
```
amuniz: It probably works here, but I saw weird things when using ```@Before```/```@After``` within ```JenkinsRule```.

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
oleg-nenashev: Thanks for catching! It's a copy-paste issue

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectPluginActionTest.java
oleg-nenashev: It would be great if you submit them as JIRA issues if these things are reproducible

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectAction.java
oleg-nenashev: Added annotations

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectPluginActionTest.java
lanwen: before and after should work fine with rule

amuniz: I don't remember exactly the under what circumstances it failed (will file the issue next time I see it. Probably it was ```@BeforeClass```/```@AfterClass```). Nevermind, if it's working here, it's working.

lanwen: @amuniz JFYI `@BeforeClass/@AfterClass` executed "inside" `@ClassRule` but "around" all `@Rule` executions.

Same with `@Before`/`@After` - they executed "inside" of `@Rule` execution (e.g. after rule starts and before rule ends)

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectPluginActionTest.java
lanwen: no, assertion error is for test failure - e.g. it should be as you expect in this case, but it not. Any unexpected failures causes test to be "broken" test. So it should be just any other exception.

oleg-nenashev: OK

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPlugin.java
```diff
@@ -0,0 +1,142 @@
...
+    void configure(EnvInjectPluginConfiguration configuration) throws IOException {
+        this.configuration = configuration;
+        VIEW_INJECTED_VARS.setEnabled(configuration.isEnablePermissions());
+        save();
+    }
+    
+    protected XmlFile getConfigXml() {
+        // We use a new configuration file, because the old one is occupied by EnvInjectNodeProperty 
+        // (ideally needs a migration of data files)
+        return new XmlFile(Jenkins.XSTREAM, new File(getJenkinsInstance().getRootDir(),"envinject2.xml"));
```
lanwen: think file name should not contain index as it very confusing. Can it be named as `envinject-global` or `envinject-plugin-configuration`?

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPlugin.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+    
+    @Override
+    public void configure(StaplerRequest req, JSONObject formData) throws IOException, ServletException, Descriptor.FormException {
+        EnvInjectPluginConfiguration _configuration = req.bindJSON(EnvInjectPluginConfiguration.class, formData);
```
lanwen: think underlines in names can be replaced with short form of the word - for example `config`. Underlines is recomended to use only for restricted java identifiers (`package`, `class`...)

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPlugin.java
oleg-nenashev: any name is confusing, because we already have envinject.xml. Probably I'll change the name to <code>org.jenkinsci.plugins.envinject.EnvInjectPlugin.xml</code> to follow the common pattern. Do you agree?

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPlugin.java
oleg-nenashev: agreed

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPlugin.java
lanwen: why not `envinject-global-plugin-configuration.xml`? With that name `envinject.xml` looks like some tmp file

oleg-nenashev: global passwords are also the global configuration :(

lanwen: if `envinject-plugin-configuration.xml` disliked too (why?), so common pattern is ok for me.

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectPluginActionTest.java
oleg-nenashev: Fixed it

lanwen: thanks!

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectAction.java
```diff
@@ -45,20 +46,38 @@ public void overrideAll(Map<String, String> all) {
         return UnmodifiableMap.decorate(envMap);
     }
 
+    @Override
```
KostyaSha: CheckForNull?

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+    
+    @Extension
+    public static final DescriptorImpl DESCRIPTOR = new DescriptorImpl();
+    
+    public static class DescriptorImpl extends Descriptor<EnvInjectPluginConfiguration> {
+
+        @Override
+        public String getDisplayName() {
+            return "N/A";
```
KostyaSha: Wrote normal name, difficult identify in descriptor list when debugging

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * Gets the default configuration of {@link EnvInjectPlugin}
+     * @return Default configuration
+     */
+    public static final @Nonnull EnvInjectPluginConfiguration getDefault() {
+        return DEFAULT;
+    }
+    
+    @Extension
+    public static final DescriptorImpl DESCRIPTOR = new DescriptorImpl();
```
KostyaSha: Why not use `@Extension` under `Descriptor` class?

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectVarList.java
```diff
@@ -125,5 +132,17 @@ private void writeJsonResponse(StaplerResponse response) throws IOException {
         outputStream.write(sb.toString().getBytes());
         outputStream.write("]}}".getBytes());
     }
+    
+    //TODO: Throw errors in responses?
+    /**
+     * Implements an {@link EnvInjectVarList}, which does not provide any variables.
+     */
+    private static class Hidden extends EnvInjectVarList {
+        private static final long serialVersionUID = 1L;
```
KostyaSha: AFAIR 1L already is for newly created classes

src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectPlugin/config.jelly
```diff
@@ -0,0 +1,34 @@
...
+-->
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" 
+         xmlns:t="/lib/hudson" xmlns:f="/lib/form" xmlns:sl="/hudson/plugins/sidebar_link">
+  <f:section title="${%Environment Injector Plugin}" description="${%Settings of the EnvInject plugin}">
+    <j:set var="instance" value="${it.configuration}" /> 
+    <j:set var="descriptor" value="${instance.descriptor}" />
+    <st:include from="${descriptor}" page="${descriptor.configPage}" optional="false" />
+  </f:section>
+</j:jelly>
```
KostyaSha: remove NL

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectVarList.java
oleg-nenashev: It's a newly created class :)

KostyaSha:
Please fix new line endings, put annotations and squash before merge, then :+1:

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectAction.java
oleg-nenashev: No need for overrides. Supermethods should do it newer APIs

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
oleg-nenashev: There was a reason to do it , but I do not recall it

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPlugin.java
```diff
@@ -0,0 +1,143 @@
...
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+import org.kohsuke.stapler.StaplerRequest;
+import org.kohsuke.stapler.export.ExportedBean;
+
+/**
+ * @author Oleg Nenashev
+ */
+@ExportedBean
+public class EnvInjectPlugin extends Plugin {
```
jglick: Using `Plugin` this way is deprecated. Use `GlobalConfiguration` instead. (Or just add a `global.jelly` view to some existing `Describable` if appropriate.)

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPlugin.java
```diff
@@ -0,0 +1,143 @@
...
+        super.load();
+        if (configuration == null) {     
+            configuration = EnvInjectPluginConfiguration.getDefault();        
+            save();
+        }
+    }
+
+    void configure(EnvInjectPluginConfiguration configuration) throws IOException {
+        this.configuration = configuration;
+        VIEW_INJECTED_VARS.setEnabled(configuration.isEnablePermissions());
```
jglick: This sucks but pending [JENKINS-17200](https://issues.jenkins-ci.org/browse/JENKINS-17200) there may be no better way.

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPlugin.java
oleg-nenashev: Yes, makes sense to migrate to GlobalConfiguration. I still like the centralized Plugin-based approach, but the most of its functionality is being covered by new extension points

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPlugin.java
oleg-nenashev: IMO all this Permission/PermissionGroup enabling/disabling stuff should be provided by Jenkins core. Keeping it on plugins level increases the entropy in any case

src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectPlugin/config.jelly
oleg-nenashev: removed the config at all after @jglick's comment

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPlugin.java
oleg-nenashev: @jglick Done

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPlugin.java
```diff
@@ -0,0 +1,112 @@
...
+        }
+        
+        // Last check - global configs
+        final EnvInjectPluginConfiguration configuration = getInstance().getConfiguration();
+        return !configuration.isHideInjectedVars();
+    }
+    
+    @Override 
+    public void start() throws Exception {
+        load();
```
lanwen: what for you call load?

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+    
+    @CheckForNull
+    public static EnvInjectPluginConfiguration getInstance() {
+        final Jenkins j = Jenkins.getInstance();
+        if (j == null) {
+            return null;
+        }
+        
+        ExtensionList<EnvInjectPluginConfiguration> extensionList =
```
lanwen: can be simplified as 
```java
EnvInjectPluginConfiguration.all().get(EnvInjectPluginConfiguration.class)
```

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPlugin.java
oleg-nenashev: Yes, no need to do it here

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPlugin.java
jglick: @oleg-nenashev agreed, which is why I think a proper solution to JENKINS-17200 is important. Not sure about the design yet. Anyway out of scope here.

oleg-nenashev:
@lanwen done

lanwen:
:+1:

src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration/help-enablePermissions.html
```diff
@@ -0,0 +1,7 @@
+<div>
+  Enables all permissions in EnvInject plugin.
+  <p/>
+  <b>Warning!</b> If you enable this option, users may lose an access to 
+  environment variables list. A manual reconfiguration of permissions will be
+  required.
+</div>
```
KostyaSha: add new lines

oleg-nenashev:
@KostyaSha done & squashed

KostyaSha:
:+1:

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPlugin.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+        
+        // Last check - global configs
+        final EnvInjectPluginConfiguration configuration = getInstance().getConfiguration();
+        return !configuration.isHideInjectedVars();
+    }
+    
+    @Override 
+    public void start() throws Exception {
+        VIEW_INJECTED_VARS.setEnabled(getConfiguration().isEnablePermissions());
```
jglick: BTW you could just add an `@Initializer` to `EnvInjectPluginConfiguration` and then kill this whole class.

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPlugin.java
```diff
@@ -0,0 +1,113 @@
...
+    }  
+    
+    //TODO: Delete after upgrading the core dependency
```
jglick: Polite to note the minimum version in such comments, for easy grepping when doing an upgrade.

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
```diff
@@ -0,0 +1,117 @@
...
+     * @return Default configuration
+     */
+    public static final @Nonnull EnvInjectPluginConfiguration getDefault() {
+        return DEFAULT;
+    }
+
+    @Override
+    protected XmlFile getConfigFile() {
+        return new XmlFile(Jenkins.XSTREAM, new File(getJenkinsInstance().getRootDir(), 
+                "envinject-plugin-configuration.xml"));
```
jglick: Unnecessary override.

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    @Override
+    public boolean configure(StaplerRequest req, JSONObject json) throws FormException {
+        final boolean newEnablePermissions = json.getBoolean("enablePermissions");
+        final boolean newHideInjectedVars = json.getBoolean("hideInjectedVars");
+        return configure(newHideInjectedVars, newEnablePermissions);
```
jglick: Probably suffices to

```java
req.bindJSON(this,json);
```

(adding `save()` to setters).

KostyaSha: Where do you see setters?

KostyaSha: bindJSON unreally difficult to debug as it contains many hacks.

src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration/config.jelly
```diff
@@ -0,0 +1,37 @@
...
+-->
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout" 
+         xmlns:t="/lib/hudson" xmlns:f="/lib/form" xmlns:sl="/hudson/plugins/sidebar_link">
+  <f:section title="${%Environment Injector Plugin}" description="${%Settings of the EnvInject plugin}">
+    <f:entry field="enablePermissions" title="${%enablePermissions.title}">
+      <f:checkbox  checked="${it.enablePermissions}"/>
+    </f:entry>
+    <f:entry field="hideInjectedVars" title="${%hideInjectedVars.title}">
+      <f:checkbox checked="${it.hideInjectedVars}"/>
```
jglick: `checked` unnecessary here—the current value is loaded according to `field`.

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectPluginActionTest.java
jglick: `@Before` does not with `RestartableJenkinsRule`, perhaps that is what you meant. Should work fine with `JenkinsRule`.

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectPluginActionTest.java
```diff
@@ -0,0 +1,153 @@
...
+        SecurityContext initialContext = null;
+        try {
+            Authentication auth = user.impersonate();
+            initialContext = hudson.security.ACL.impersonate(auth);
+            return EnvInjectPlugin.canViewInjectedVars(run);
+        } finally {
+            if (initialContext != null) {
+                SecurityContextHolder.setContext(initialContext);
+            }
+        }
```
jglick: BTW can be written more concisely with an effectively final variable:

```java
Authentication auth = user.impersonate();
SecurityContext initialContext = ACL.impersonate(auth);
try {
    return EnvInjectPlugin.canViewInjectedVars(run);
} finally {
    SecurityContextHolder.setContext(initialContext);
}
```

(There are also two newer overloads with closures, but neither are suitable for Java 8 lambdas.)

jglick:
:bee:

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
oleg-nenashev: <code>bindJSON()</code> didn't work for me, and I had to use a straightforward approach (may depend on the Jenkins core version)

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
oleg-nenashev: The plugin declares envinject.xml in EnvInjectNodeProperty (which is a problematic approach BTW), so I finally decided to create a new config file. Probably <code>GlobalConfiguration</code> sets another default value

jglick: It does.

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
jglick: > `bindJSON()` didn't work for me

Probably because you forgot the setters.

KostyaSha: Just because of bindJSON you must create setters with save() in each.

jglick: > Just because of bindJSON you must create setters with save() in each.

This is proper style for a configuration object anyway. Useful from Groovy scripting, etc.

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
```diff
@@ -0,0 +1,117 @@
...
+    private boolean hideInjectedVars;
+
+    private boolean enablePermissions;
+
+    public EnvInjectPluginConfiguration() {
+        load();
+    }
+    
+    @DataBoundConstructor
+    public EnvInjectPluginConfiguration(boolean hideInjectedVars, boolean enablePermissions) {
```
jglick: AFAIK this will never be called and should be removed. It is not a `Describable` that may be instantiated, it is a singleton.

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
KostyaSha: and then synchronise everything, useless complexity for 2 fields.

jglick: Synchronization would be overkill.

I do not really care how you choose to implement, just noting the normal pattern used in contemporary `GlobalConfiguration` subclasses FYI.

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPlugin.java
oleg-nenashev: I still prefer to keep permissions in the XYZPlugin class instead of its configuration

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
lanwen: and how works unmarshalling then?

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
lanwen: bindJson works fine here: https://github.com/jenkinsci/github-plugin/pull/68/files#diff-0cd2e303488033a692923bea444e85ecR155

lanwen: This works fine https://github.com/jenkinsci/github-plugin/pull/68/files#diff-0cd2e303488033a692923bea444e85ecR160

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
oleg-nenashev: unmarshalling works via <code>load()</code>, which is being called by the constructor during the singleton instatination. Seems DatatBoundConstructor should be removed

src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration/config.jelly
oleg-nenashev: IIRC there were issues in old Jenkins cores. BTW, I can check & cleanup it

oleg-nenashev:
retriggering the build to get the artifact

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
jglick: > what about fact that GlobalConfig is Describable?

I guess it needs an associated `Descriptor` for display purposes.

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
oleg-nenashev: IIRC @lanwen was against the approach with a full package/class name. I also agree with it since we have "envinject.xml". It makes sense to merge configurations to a common name later in envinject-2.0

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
oleg-nenashev: Probably it's version-dependant. I'm against setters, so let's leave the implementation as is till we bump the core dependency.

src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration/config.jelly
oleg-nenashev: Yes, works well wit the current baseline

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectPluginActionTest.java
oleg-nenashev: fixed it

oleg-nenashev:
@reviewbybees done
I've fixed the minor code quality issues discovered by @jglick + commented on the config location topic. Since I had a bee from him before that, I'm using my maintainer's role to go forward and to release the 0.92-beta-1 version. There will be a short timeframe before the full release in order to collect the feedback, so I can process additional feedback if it appears.

recena:
@oleg-nenashev You should add your info in the `developers` section (`pom.file`) if you are going to be maintainer.

oleg-nenashev:
@recena done

PR #58, opened 2015-08-11
Add new tests for EnvInjectJobProperty

oleg-nenashev:
One of tests is failing and indicate the real issue, which should be fixed IMO.

@reviewbybees @recena @lanwen

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -89,7 +89,12 @@
             <version>${mockito.version}</version>
             <scope>test</scope>
         </dependency>
-
+        <dependency>
+            <groupId>org.jenkins-ci</groupId>
+            <artifactId>test-annotations</artifactId>
+            <version>1.2</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <repositories>
--- a/src/test/java/org/jenkinsci/plugins/envinject/EnvInjectJobPropertyTest.java
+++ b/src/test/java/org/jenkinsci/plugins/envinject/EnvInjectJobPropertyTest.java
@@ -1,0 +1,130 @@
+package org.jenkinsci.plugins.envinject;
+
+import hudson.model.Cause;
+import hudson.model.FreeStyleBuild;
+import hudson.model.FreeStyleProject;
+import hudson.model.ParametersAction;
+import hudson.model.StringParameterValue;
+import hudson.model.TaskListener;
+import hudson.model.queue.QueueTaskFuture;
+import java.io.IOException;
+import javax.annotation.Nonnull;
+import org.junit.Rule;
+import org.jvnet.hudson.test.JenkinsRule;
+
+import static org.junit.Assert.*;
+import org.junit.Ignore;
+import org.junit.Test;
+import org.jvnet.hudson.test.CaptureEnvironmentBuilder;
+import org.jvnet.hudson.test.Issue;
+
+/**
+ * Tests for {@link EnvInjectJobProperty}
+ * @author Oleg Nenashev
+ */
+public class EnvInjectJobPropertyTest {
+    
+    @Rule
+    public JenkinsRule jenkinsRule = new JenkinsRule();
+
+    @Test
+    public void shouldInjectBuildVarsFromPropertiesContent() throws Exception {   
+        FreeStyleProject project = jenkinsRule.createFreeStyleProject();
+        EnvInjectJobProperty<FreeStyleProject> prop = forPropertiesContent(project, "FOO=BAR");
+        
+        FreeStyleBuild build = jenkinsRule.buildAndAssertSuccess(project);
+        assertEquals("The FOO variable has not been injected properly", "BAR", 
+                build.getEnvironment(TaskListener.NULL).get("FOO"));
+    }
+    
+    @Test
+    public void shouldNotInjectVariablesIfPropertyIsDisabled() throws Exception {   
+        FreeStyleProject project = jenkinsRule.createFreeStyleProject();
+        
+        EnvInjectJobProperty<FreeStyleProject> prop = new EnvInjectJobProperty<FreeStyleProject>();
+        prop.setInfo(new EnvInjectJobPropertyInfo(null, "FOO=BAR", null, null, null, false));
+        // prop.setOn(false); // It is default
+        project.addProperty(prop);
+        
+        FreeStyleBuild build = jenkinsRule.buildAndAssertSuccess(project);
+        assertNull("The plugin should not inject properties by default", 
+                build.getEnvironment(TaskListener.NULL).get("FOO"));
+    }
+    
+    @Test
+    public void shouldKeepBuildVariablesByDefault() throws Exception {   
+        FreeStyleProject project = jenkinsRule.createFreeStyleProject();     
+        // We assign a value to another parameter just to enable the engine
+        EnvInjectJobProperty<FreeStyleProject> prop = forPropertiesContent(project, "PARAM2=Overridden");
+        
+        QueueTaskFuture<FreeStyleBuild> scheduled = project.scheduleBuild2(0, new Cause.UserIdCause(),
+                new ParametersAction(new StringParameterValue("PARAM", "Value")));
+        assertNotNull(scheduled);
+        FreeStyleBuild build = scheduled.get();
+        jenkinsRule.assertBuildStatusSuccess(build);
+        assertEquals("The build parameter has been overriden", "Value", 
+                build.getEnvironment(TaskListener.NULL).get("PARAM"));
+    }
+    
+    @Test
+    @Ignore("The value is being actually contributed by other env contributors. The feature seems to be obsolete")
+    public void shouldNotKeepBuildVariablesIfDisabled() throws Exception {   
+        FreeStyleProject project = jenkinsRule.createFreeStyleProject();
+        
+        EnvInjectJobProperty<FreeStyleProject> prop = forPropertiesContent(project, "PARAM2=Overridden");
+        prop.setKeepBuildVariables(false);
+        
+        QueueTaskFuture<FreeStyleBuild> scheduled = project.scheduleBuild2(0, new Cause.UserIdCause(),
+                new ParametersAction(new StringParameterValue("PARAM", "Value")));
+        assertNotNull(scheduled);
+        FreeStyleBuild build = scheduled.get();
+        jenkinsRule.assertBuildStatusSuccess(build);
+        assertNull("We expect that the PARAM is not specified", 
+                build.getEnvironment(TaskListener.NULL).get("PARAM"));
+    }
+    
+    @Test
+    @Ignore("It should not override vars accoroding to the manual testing. But it does... "
+            + "Manual tests also show the wrong value in InjectedVarsAction")
+    @Issue("JENKINS-29905")
+    public void shouldNotOverrideBuildParametersByDefault() throws Exception {   
+        FreeStyleProject project = jenkinsRule.createFreeStyleProject();    
+        EnvInjectJobProperty<FreeStyleProject> prop = forPropertiesContent(project, "PARAM=Overridden");
+        prop.setOverrideBuildParameters(false);
+        
+        final CaptureEnvironmentBuilder envCapture = new CaptureEnvironmentBuilder();
+        project.getBuildersList().add(envCapture);
+        
+        QueueTaskFuture<FreeStyleBuild> scheduled = project.scheduleBuild2(0, new Cause.UserIdCause(),
+                new ParametersAction(new StringParameterValue("PARAM", "ValueFromParameter")));
+        assertNotNull(scheduled);
+        FreeStyleBuild build = scheduled.get();
+        jenkinsRule.assertBuildStatusSuccess(build);
+        assertEquals("The variable has been overriden in the environment", "ValueFromParameter", envCapture.getEnvVars().get("PARAM"));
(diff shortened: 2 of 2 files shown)


src/test/java/org/jenkinsci/plugins/envinject/EnvInjectJobPropertyTest.java
```diff
@@ -0,0 +1,134 @@
...
+        assertNotNull(scheduled);
+        FreeStyleBuild build = scheduled.get();
+        assertEquals(Result.SUCCESS, build.getResult());
+        assertNull("We expect that the PARAM is not specified", build.getEnvironment(TaskListener.NULL).get("PARAM"));
+    }
+    
+    @Test
+    @Ignore("It should not override vars accroding to the manual testing. But it does... "
+            + "Manual tests also show the wrong value in InjectedVarsAction")
+    public void testOverrideBuildParameters_Default() throws Exception {
```
oleg-nenashev: @recena This is an issue I've discovered in https://github.com/jenkinsci/envinject-plugin/pull/56 . After an investigation it's not related to your PR

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectJobPropertyTest.java
```diff
@@ -0,0 +1,134 @@
...
+        QueueTaskFuture<FreeStyleBuild> scheduled = project.scheduleBuild2(0, new Cause.UserIdCause(),
+                new ParametersAction(new StringParameterValue("PARAM", "Value")));
+        assertNotNull(scheduled);
+        FreeStyleBuild build = scheduled.get();
+        assertEquals(Result.SUCCESS, build.getResult());
+        assertNull("We expect that the PARAM is not specified", build.getEnvironment(TaskListener.NULL).get("PARAM"));
+    }
+    
+    @Test
+    @Ignore("It should not override vars accroding to the manual testing. But it does... "
```
jglick: typo

jglick:
:bee:, it is always nice to have some baseline tests.

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectJobPropertyTest.java
```diff
@@ -0,0 +1,134 @@
...
+ * Tests for {@link EnvInjectJobProperty}
+ * @author Oleg Nenashev
+ */
+public class EnvInjectJobPropertyTest {
+    
+    @Rule
+    public JenkinsRule j = new JenkinsRule();
+
+    @Test
+    public void testVariableInjection() throws Exception {
```
lanwen: please name tests as `should` + what we are expect + `When` + we smth do.
The "test" prefix is from junit3 and now this role plays `@Test` annotation.

so for example here - `shouldInjectVarsWhenContributeFromJobProperty`

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectJobPropertyTest.java
```diff
@@ -0,0 +1,134 @@
...
+        prop.setInfo(new EnvInjectJobPropertyInfo(null, "FOO=BAR", null, null, null, false));
+        prop.setOn(true);
+        project.addProperty(prop);
+        
+        FreeStyleBuild build = j.buildAndAssertSuccess(project);
+        assertEquals("BAR", build.getEnvironment(TaskListener.NULL).get("FOO"));
+    }
+    
+    @Test
+    public void testVariableInjection_PropertyDisabled() throws Exception {
```
lanwen: `shouldNotInjectVarsFromDisabledJobProp`

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectJobPropertyTest.java
```diff
@@ -0,0 +1,134 @@
...
+        FreeStyleBuild build = j.buildAndAssertSuccess(project);
+        assertEquals("BAR", build.getEnvironment(TaskListener.NULL).get("FOO"));
+    }
+    
+    @Test
+    public void testVariableInjection_PropertyDisabled() throws Exception {   
+        FreeStyleProject project = j.createFreeStyleProject();
+        
+        EnvInjectJobProperty<FreeStyleProject> prop = new EnvInjectJobProperty<FreeStyleProject>();
+        prop.setInfo(new EnvInjectJobPropertyInfo(null, "FOO=BAR", null, null, null, false));
```
lanwen: what does mean all this nulls and false? Think the only useful thing is var value. So this two lines can be replaced with one method:
```java
public EnvInjectJobProperty prop(String content) {
       EnvInjectJobProperty<FreeStyleProject> prop = new EnvInjectJobProperty<FreeStyleProject>();
       prop.setInfo(new EnvInjectJobPropertyInfo(null, content, null, null, null, false));
}
```

so it simplifies all tests

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectJobPropertyTest.java
```diff
@@ -0,0 +1,134 @@
...
+        
+        EnvInjectJobProperty<FreeStyleProject> prop = new EnvInjectJobProperty<FreeStyleProject>();
+        prop.setOn(true);
+        prop.setInfo(new EnvInjectJobPropertyInfo(null, null, null, null, null, false));
+        // Default keepBuildVariables should be true
+        project.addProperty(prop);
+        
+        QueueTaskFuture<FreeStyleBuild> scheduled = project.scheduleBuild2(0, new Cause.UserIdCause(),
+                new ParametersAction(new StringParameterValue("PARAM", "Value")));
+        assertNotNull(scheduled);
```
lanwen: please add a little comment for asserts. In standard log its difficult to understand cause of the fail if you have only stack-trace

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectJobPropertyTest.java
```diff
@@ -0,0 +1,134 @@
...
+        FreeStyleBuild build = scheduled.get();
+        assertEquals(Result.SUCCESS, build.getResult());
```
lanwen: please don't use `assertEquals`. Use `assertThat(actual, equalTo(expected))` as of the majority of people don't remember where the expected and where the actual should be in assertEquals. It causes sometimes very confusing assert messages.

Also, assertThat is more maintain-friendly as better refactorable solution (you need to change only matcher, and not whole statement to change logic)

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectJobPropertyTest.java
```diff
@@ -0,0 +1,134 @@
...
+import org.jvnet.hudson.test.CaptureEnvironmentBuilder;
+
+/**
+ * Tests for {@link EnvInjectJobProperty}
+ * @author Oleg Nenashev
+ */
+public class EnvInjectJobPropertyTest {
+    
+    @Rule
+    public JenkinsRule j = new JenkinsRule();
```
lanwen: please don't name any var with one symbol. Even if it test field. (Exclusions only i,j for loops). `jenkins` is not too long to avoid writing it

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectJobPropertyTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+    @Test
+    public void testVariableInjection() throws Exception {   
+        FreeStyleProject project = j.createFreeStyleProject();
```
lanwen: move such repeated lines to field with init  in`@Before`. (This simple refactoring saves 6 lines here with cost of 4, not bad)

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectJobPropertyTest.java
jglick: `jenkins` would be a poor name here as it could be confused with the field of that name in `JenkinsRule`.

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectJobPropertyTest.java
oleg-nenashev: IIRC it's a common pattern for tests, which imply the parameter variation, but do not implement them via withParameters tag

pom.xml
```diff
@@ -89,7 +89,13 @@
             <version>${mockito.version}</version>
             <scope>test</scope>
         </dependency>
-
+        <dependency>
+            <groupId>org.jenkins-ci</groupId>
+            <artifactId>test-annotations</artifactId>
+            <version>1.2</version>
+            <scope>test</scope>
+            <type>jar</type>
```
lanwen: it is jar by default. Why you specify it here?

oleg-nenashev: netbeans injected it automatically. I'll remove it

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectJobPropertyTest.java
```diff
@@ -0,0 +1,130 @@
...
+        FreeStyleProject project = jenkinsRule.createFreeStyleProject();
+        
+        EnvInjectJobProperty<FreeStyleProject> prop = forPropertiesContent(project, "PARAM2=Overridden");
+        prop.setKeepBuildVariables(false);
+        
+        QueueTaskFuture<FreeStyleBuild> scheduled = project.scheduleBuild2(0, new Cause.UserIdCause(),
+                new ParametersAction(new StringParameterValue("PARAM", "Value")));
+        assertNotNull(scheduled);
+        FreeStyleBuild build = scheduled.get();
+        jenkinsRule.assertBuildStatusSuccess(build);
```
jglick: A cumbersome way of writing

```java
FreeStyleBuild build = jenkinsRule.assertBuildStatusSuccess(project.scheduleBuild2(0, new Cause.UserIdCause(),
    new ParametersAction(new StringParameterValue("PARAM", "Value")));
```

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectJobPropertyTest.java
```diff
@@ -0,0 +1,130 @@
...
+        final CaptureEnvironmentBuilder envCapture = new CaptureEnvironmentBuilder();
+        project.getBuildersList().add(envCapture);
+        
+        QueueTaskFuture<FreeStyleBuild> scheduled = project.scheduleBuild2(0, new Cause.UserIdCause(),
+                new ParametersAction(new StringParameterValue("PARAM", "ValueFromParameter")));
+        assertNotNull(scheduled);
+        FreeStyleBuild build = scheduled.get();
+        jenkinsRule.assertBuildStatusSuccess(build);
+        assertEquals("The variable has been overriden in the environment", "ValueFromParameter", envCapture.getEnvVars().get("PARAM"));
+        assertEquals("The variable has been overriden in the API", "ValueFromParameter", build.getEnvironment(TaskListener.NULL).get("PARAM"));
```
jglick: `StreamTaskListener.fromStdout`

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectJobPropertyTest.java
```diff
@@ -0,0 +1,130 @@
...
+    
+    @Test
+    @Ignore("The value is being actually contributed by other env contributors. The feature seems to be obsolete")
+    public void shouldNotKeepBuildVariablesIfDisabled() throws Exception {   
+        FreeStyleProject project = jenkinsRule.createFreeStyleProject();
+        
+        EnvInjectJobProperty<FreeStyleProject> prop = forPropertiesContent(project, "PARAM2=Overridden");
+        prop.setKeepBuildVariables(false);
+        
+        QueueTaskFuture<FreeStyleBuild> scheduled = project.scheduleBuild2(0, new Cause.UserIdCause(),
```
jglick: You can omit `new Cause.UserIdCause()` (just pass null).

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectJobPropertyTest.java
```diff
@@ -0,0 +1,130 @@
...
+                new ParametersAction(new StringParameterValue("PARAM", "Value")));
+        assertNotNull(scheduled);
+        FreeStyleBuild build = scheduled.get();
+        jenkinsRule.assertBuildStatusSuccess(build);
+        assertNull("We expect that the PARAM is not specified", 
+                build.getEnvironment(TaskListener.NULL).get("PARAM"));
+    }
+    
+    @Test
+    @Ignore("It should not override vars accoroding to the manual testing. But it does... "
```
jglick: typo

jglick:
:bee:

lanwen:
:+1:

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectJobPropertyTest.java
oleg-nenashev: crap. I've replaced one typo by another one :(

oleg-nenashev:
@jglick 
I've added the tests polishing to my TODO list in order to revise them after the release. 

@reviewbybees done

PR #71, opened 2015-09-09
Fixes JENKINS-23274: remove log of script content

jamestelfer:
Avoids the script content appearing in the build output entirely. This is the simplest possible change to fix this issue, but it also seems the most suitable.

This probably isn't the most high value issue against the plugin, but it would clear up my build logs so it would be nice. Thoughts?

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectEnvVars.java
+++ b/src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectEnvVars.java
@@ -121,7 +121,7 @@ public class EnvInjectEnvVars implements Serializable {
             return new HashMap<String, String>();
         }
 
-        logger.info(String.format("Evaluation the following Groovy script content: %n%s%n", scriptContent));
+        logger.info("Evaluating the Groovy script content");
 
         Binding binding = new Binding();
         String jobName = envVars.get("JOB_NAME");


martinda:
Thank you. I find this problem excessively annoying. Leaking the job configuration to the console on a production system is bad practice, and possibly a security concern for some.

oleg-nenashev:
IMO the logging target and level should be changed, but the complete removal is not a good approach:

My proposals:
* Print "Evaluating the Groovy script content" to the current log (w/o contents)
* Print the contents of the Groovy Script to the <code>java.util.logging LOGGER</code> with the CONFIG logging level

jamestelfer:
After comments from @oleg-nenashev, I've reinstated the log statement. It just logs that the script is being executed though, not the content.

I didn't add the `java.util.logging` statement, though, because:
1. The plugin doesn't appear to use this anywhere else, and
2. AFAICT the plugin doesn't actively seek to tweak the script itself, just executes what's in the configuration.

Let me know if this reasoning doesn't work for you.

oleg-nenashev:
Hi James. I agree. +1 for the merge

 Cc'ed @recena to get a second sign-off.

recena:
:+1:

oleg-nenashev:
@jamestelfer could you please squash the commit (and probably to remove the empty line)? After that I'll merge the PR. Not sure about the release date, probably we'll try to merge in several other pull requests on this week.

jamestelfer:
After a false start, I've squashed the commit. Thanks for walking me through the process, even though it was such a minor change.

oleg-nenashev:
Merged, thanks a lot for your contribution!

jamestelfer:
Simple fix I know but I'm pretty stoked. Thanks for taking the time to push it through!

t-stamm:
Any plans to release this in the near future?

PR #72, opened 2015-09-11
Add "USER_NAME"  to BuildVariables

zhangzhenyu0819:
Add "USER_NAME"  to BuildVariables(ManualBuildCause only)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/envinject/service/BuildCauseRetriever.java
+++ b/src/main/java/org/jenkinsci/plugins/envinject/service/BuildCauseRetriever.java
@@ -51,26 +51,22 @@ public class BuildCauseRetriever {
     public String getCauseUserName(AbstractBuild<?, ?> build) {
         CauseAction causeAction = build.getAction(CauseAction.class);
         List<Cause> buildCauses = causeAction.getCauses();
-
         for (Cause cause : buildCauses) {
             if (isUserCause(cause) || isUserIdCause(cause)) {
             	return getUserName(cause);
             }
         }
-
 		return null;
 	}
 
     public String getCauseUserId(AbstractBuild<?, ?> build) {
         CauseAction causeAction = build.getAction(CauseAction.class);
         List<Cause> buildCauses = causeAction.getCauses();
-
         for (Cause cause : buildCauses) {
             if (isUserIdCause(cause)) {
             	return getUserId(cause);
             }
         }
-
 		return null;
 	}
 
@@ -126,7 +122,6 @@ public class BuildCauseRetriever {
     	if (cause instanceof UserCause){
     		return true;
     	}
-
     	return false;
     }
 
@@ -134,7 +129,6 @@ public class BuildCauseRetriever {
     	if (cause instanceof UserIdCause){
     		return true;
     	}
-
     	return false;
     }
 
@@ -143,11 +137,10 @@ public class BuildCauseRetriever {
     	if (cause instanceof UserIdCause) {
     		Cause.UserIdCause userIdCause = (UserIdCause) cause;
     		return userIdCause.getUserName();
-        }else if (cause instanceof UserCause) {
+        } else if (cause instanceof UserCause) {
     		Cause.UserCause userCause = (UserCause) cause;
     		return userCause.getUserName();
         }
-
         return null;
     }
 
@@ -157,7 +150,6 @@ public class BuildCauseRetriever {
     		Cause.UserIdCause userIdCause = (UserIdCause) cause;
     		return userIdCause.getUserId();
         }
-
         return null;
     }
 
--- a/src/main/java/org/jenkinsci/plugins/envinject/service/BuildCauseRetriever.java
+++ b/src/main/java/org/jenkinsci/plugins/envinject/service/BuildCauseRetriever.java
@@ -53,8 +53,21 @@ public class BuildCauseRetriever {
         List<Cause> buildCauses = causeAction.getCauses();
 
         for (Cause cause : buildCauses) {
-            if (isUserCause(cause)) {
+            if (isUserCause(cause) || isUserIdCause(cause)) {
             	return getUserName(cause);
+            }
+        }
+
+		return null;
+	}
+
+    public String getCauseUserId(AbstractBuild<?, ?> build) {
+        CauseAction causeAction = build.getAction(CauseAction.class);
+        List<Cause> buildCauses = causeAction.getCauses();
+
+        for (Cause cause : buildCauses) {
+            if (isUserIdCause(cause)) {
+            	return getUserId(cause);
             }
         }
 
@@ -110,23 +123,39 @@ public class BuildCauseRetriever {
     }
 
     private static Boolean isUserCause(Cause cause) {
-    	if (Cause.UserIdCause.class.isInstance(cause)) {
+    	if (cause instanceof UserCause){
     		return true;
-        }else if (Cause.UserCause.class.isInstance(cause)) {
+    	}
+
+    	return false;
+    }
+
+    private static Boolean isUserIdCause(Cause cause) {
+    	if (cause instanceof UserIdCause){
     		return true;
-        }
+    	}
 
     	return false;
     }
 
     @CheckForNull
     private static String getUserName(Cause cause) {
-    	if (Cause.UserIdCause.class.isInstance(cause)) {
+    	if (cause instanceof UserIdCause) {
     		Cause.UserIdCause userIdCause = (UserIdCause) cause;
     		return userIdCause.getUserName();
-        }else if (Cause.UserCause.class.isInstance(cause)) {
-    		Cause.UserCause userIdCause = (UserCause) cause;
-    		return userIdCause.getUserName();
+        }else if (cause instanceof UserCause) {
+    		Cause.UserCause userCause = (UserCause) cause;
+    		return userCause.getUserName();
+        }
+
+        return null;
+    }
+
+    @CheckForNull
+    private static String getUserId(Cause cause) {
+    	if (cause instanceof UserIdCause) {
+    		Cause.UserIdCause userIdCause = (UserIdCause) cause;
+    		return userIdCause.getUserId();
         }
 
         return null;
--- a/src/main/java/org/jenkinsci/plugins/envinject/service/BuildCauseRetriever.java
+++ b/src/main/java/org/jenkinsci/plugins/envinject/service/BuildCauseRetriever.java
@@ -1,9 +1,10 @@
 package org.jenkinsci.plugins.envinject.service;
 
-import static com.google.common.base.Joiner.*;
-import static org.apache.commons.lang.StringUtils.*;
+import static com.google.common.base.Joiner.on;
+import static org.apache.commons.lang.StringUtils.isNotBlank;
 import hudson.model.AbstractBuild;
 import hudson.model.Cause;
+import hudson.model.Cause.UserCause;
 import hudson.model.Cause.UserIdCause;
 import hudson.model.CauseAction;
 import hudson.triggers.SCMTrigger;
@@ -17,6 +18,8 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+
+import edu.umd.cs.findbugs.annotations.CheckForNull;
 
 
 /**
@@ -51,13 +54,12 @@ public class BuildCauseRetriever {
         List<Cause> buildCauses = causeAction.getCauses();
 
         for (Cause cause : buildCauses) {
-            if (isUserCause(cause)){
+            if (isUserCause(cause)) {
             	return getUserName(cause);
             }
         }
 
 		return null;
-
 	}
 
     private static void insertRootCauseNames(Set<String> causeNames, Cause cause, int depth) {
@@ -111,14 +113,20 @@ public class BuildCauseRetriever {
     private static Boolean isUserCause(Cause cause) {
     	if (Cause.UserIdCause.class.isInstance(cause)) {
     		return true;
+        }else if (Cause.UserCause.class.isInstance(cause)) {
+    		return true;
         }
 
     	return false;
     }
 
+    @CheckForNull
     private static String getUserName(Cause cause) {
     	if (Cause.UserIdCause.class.isInstance(cause)) {
     		Cause.UserIdCause userIdCause = (UserIdCause) cause;
+    		return userIdCause.getUserName();
+        }else if (Cause.UserCause.class.isInstance(cause)) {
+    		Cause.UserCause userIdCause = (UserCause) cause;
     		return userIdCause.getUserName();
         }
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/envinject/service/BuildCauseRetriever.java
```diff
@@ -46,6 +46,20 @@
...
+    public String getCauseUserName(AbstractBuild<?, ?> build) {
+        CauseAction causeAction = build.getAction(CauseAction.class);
+        List<Cause> buildCauses = causeAction.getCauses();
+
+        for (Cause cause : buildCauses) {
+            if (isUserCause(cause)){
+            	return getUserName(cause);
+            }
+        }
+
```
lanwen: redundant line breaks

src/test/java/org/jenkinsci/plugins/envinject/BuildCauseRetrieverTest.java
```diff
@@ -1,24 +1,24 @@
 package org.jenkinsci.plugins.envinject;
 
+import static com.google.common.base.Joiner.*;
```
lanwen: please dont use wildcard imports. They hamper to see which classes is actually imported

src/test/java/org/jenkinsci/plugins/envinject/BuildCauseRetrieverTest.java
```diff
@@ -39,13 +40,14 @@
     @SuppressWarnings("deprecation")
     @Test
     public void shouldWriteInfoAboutManualBuildCause() throws Exception {
```
lanwen: please add test for not null value of username for user cause

src/main/java/org/jenkinsci/plugins/envinject/service/BuildCauseRetriever.java
```diff
@@ -94,4 +108,21 @@ private static String getTriggerName(Cause cause) {
         return null;
     }
 
+    private static Boolean isUserCause(Cause cause) {
+    	if (Cause.UserIdCause.class.isInstance(cause)) {
+    		return true;
+        }
```
lanwen: please format all changed code properly

src/main/java/org/jenkinsci/plugins/envinject/service/BuildCauseRetriever.java
```diff
@@ -94,4 +108,21 @@ private static String getTriggerName(Cause cause) {
         return null;
     }
 
+    private static Boolean isUserCause(Cause cause) {
+    	if (Cause.UserIdCause.class.isInstance(cause)) {
```
oleg-nenashev: This code should also support the deprecated <code>UserCause</code>, because it's still being used in Hudson-compatible plugins

oleg-nenashev: Also use <code>cause instanceof Cause.UserIdCause</code>

src/main/java/org/jenkinsci/plugins/envinject/service/BuildCauseRetriever.java
```diff
@@ -94,4 +108,21 @@ private static String getTriggerName(Cause cause) {
...
+    		return true;
+        }
+
+    	return false;
+    }
+
+    private static String getUserName(Cause cause) {
+    	if (Cause.UserIdCause.class.isInstance(cause)) {
```
oleg-nenashev: same as above (legacy class and Java operator)

src/main/java/org/jenkinsci/plugins/envinject/service/BuildCauseRetriever.java
```diff
@@ -94,4 +108,21 @@ private static String getTriggerName(Cause cause) {
...
 
+    private static Boolean isUserCause(Cause cause) {
+    	if (Cause.UserIdCause.class.isInstance(cause)) {
+    		return true;
+        }
+
+    	return false;
+    }
+
+    private static String getUserName(Cause cause) {
```
oleg-nenashev: Add <code>@CheckForNull</code>, because the method may return null

src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectVariableGetter.java
```diff
@@ -3,14 +3,17 @@
 import hudson.EnvVars;
 import hudson.Util;
 import hudson.matrix.MatrixRun;
-import hudson.model.*;
+import hudson.model.Environment;
+import hudson.model.AbstractBuild;
+import hudson.model.Computer;
+import hudson.model.EnvironmentContributor;
+import hudson.model.Executor;
+import hudson.model.Hudson;
```
oleg-nenashev: I agree with this refactoring, but it should be done in a separate pull-request

src/test/java/org/jenkinsci/plugins/envinject/BuildCauseRetrieverTest.java
```diff
@@ -39,13 +40,14 @@
...
-        Cause cause = Cause.UserCause.class.newInstance();
+    	UserIdCause cause = Cause.UserIdCause.class.newInstance();
```
oleg-nenashev: I just wonder how it may work, because you don't specify the username when you trigger the build... If the check below works, then something is wrong with the matcher

lanwen: @oleg-nenashev matcher works fine. Both values are null. So it equal

zhangzhenyu0819:
@lanwen @oleg-nenashev Thank you for your comment.
I've modified the code.

src/main/java/org/jenkinsci/plugins/envinject/service/BuildCauseRetriever.java
```diff
@@ -46,6 +48,32 @@
...
+        CauseAction causeAction = build.getAction(CauseAction.class);
+        List<Cause> buildCauses = causeAction.getCauses();
+
+        for (Cause cause : buildCauses) {
+            if (isUserCause(cause) || isUserIdCause(cause)) {
+            	return getUserName(cause);
+            }
+        }
+
+		return null;
```
lanwen: reformat this please and rm prepending empty line

src/main/java/org/jenkinsci/plugins/envinject/service/BuildCauseRetriever.java
```diff
@@ -94,4 +122,43 @@ private static String getTriggerName(Cause cause) {
...
+
+    	return false;
+    }
+
+    @CheckForNull
+    private static String getUserName(Cause cause) {
+    	if (cause instanceof UserIdCause) {
+    		Cause.UserIdCause userIdCause = (UserIdCause) cause;
+    		return userIdCause.getUserName();
+        }else if (cause instanceof UserCause) {
```
lanwen: `} else`

src/test/java/org/jenkinsci/plugins/envinject/BuildCauseRetrieverTest.java
```diff
@@ -39,13 +41,19 @@
...
         FreeStyleBuild build = jenkins.createFreeStyleProject().scheduleBuild2(0, cause).get();
 
         assertThat(build.getResult(), is(SUCCESS));
         assertThat(build, withCause(BUILD_CAUSE, MANUAL_TRIGGER));
         assertThat(build, withCause(ROOT_BUILD_CAUSE, MANUAL_TRIGGER));
         assertThat(build, withCausesIsTrue(sub(BUILD_CAUSE, MANUAL_TRIGGER), sub(ROOT_BUILD_CAUSE, MANUAL_TRIGGER)));
+        if(cause.getUserName() != null){
```
lanwen: no way. It should be separate test case

lanwen:
What for you write both `user_id` and `user_name` as of it synonyms in such context?

src/main/java/org/jenkinsci/plugins/envinject/service/BuildCauseRetriever.java
zhangzhenyu0819: I have fix here.

src/main/java/org/jenkinsci/plugins/envinject/service/BuildCauseRetriever.java
zhangzhenyu0819: i'm sorry.. I have fix here.

src/test/java/org/jenkinsci/plugins/envinject/BuildCauseRetrieverTest.java
zhangzhenyu0819: added testcase.

zhangzhenyu0819:
Additional items are being used in e-mail notification template.
e.g.
```
user_id:${USER_ID} user_name:${USER_NAME} Start this job.
```

src/test/java/org/jenkinsci/plugins/envinject/BuildCauseRetrieverTest.java
```diff
@@ -139,6 +142,32 @@ public String getShortDescription() {
...
+        assertThat(build.getResult(), is(SUCCESS));
+        assertThat(build, withCause(BUILD_CAUSE, MANUAL_TRIGGER));
+        assertThat(build, withCause(ROOT_BUILD_CAUSE, MANUAL_TRIGGER));
+        assertThat(build, withCausesIsTrue(sub(BUILD_CAUSE, MANUAL_TRIGGER), sub(ROOT_BUILD_CAUSE, MANUAL_TRIGGER)));
+        assertThat(build, withCause(USER_NAME, cause.getUserName()));
+    }
+
+    @Test
+    public void shouldWriteInfoUserIdCause() throws Exception {
+    	UserIdCause cause = Cause.UserIdCause.class.newInstance();
```
lanwen: and where user_id is set to check it? You just copied prev testcase. Unit test should test only what you want to check. Please add separate cases for nonnull/null user_id and user_name only. Don't touch other cases if they works fine.

lanwen:
@zhangzhenyu0819 is you need both user_id and user_name at same time? It adds a lot of duplication in this implementation.

zhangzhenyu0819:
@lanwen I wishes to obtain all of the user information that can be achieved, including user_id and user_name.

src/test/java/org/jenkinsci/plugins/envinject/BuildCauseRetrieverTest.java
```diff
@@ -139,6 +142,25 @@ public String getShortDescription() {
         assertThat(build, withCause(ROOT_BUILD_CAUSE, ""));
     }
 
+    @SuppressWarnings("deprecation")
+    @Test
+    public void shouldWriteInfoUserCause() throws Exception {
+        UserCause cause = Cause.UserCause.class.newInstance();
```
lanwen: when you create cause in such way, username and user id is null. Please specify it explicitly

zhangzhenyu0819:
@lanwen I have modified the test case, please check it.

src/main/java/org/jenkinsci/plugins/envinject/service/BuildCauseRetriever.java
```diff
@@ -94,4 +118,39 @@ private static String getTriggerName(Cause cause) {
...
+        } else if (cause instanceof UserCause) {
+    		Cause.UserCause userCause = (UserCause) cause;
+    		return userCause.getUserName();
+        }
+        return null;
+    }
+
+    @CheckForNull
+    private static String getUserId(Cause cause) {
+    	if (cause instanceof UserIdCause) {
```
oleg-nenashev: Missing UserCause handling

lanwen:
@zhangzhenyu0819 thanks, now tests is ok

src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectVariableGetter.java
```diff
@@ -111,6 +112,20 @@
         Map<String, String> triggerVariable = new BuildCauseRetriever().getTriggeredCause(build);
         result.putAll(triggerVariable);
 
+        //Get Cause UserName
+        String userName = new BuildCauseRetriever().getCauseUserName(build);
```
oleg-nenashev: I would refactor the code to call <code>new BuildCauseRetriever()</code> only once in order to simplify the life of the garbage collector

src/main/java/org/jenkinsci/plugins/envinject/service/BuildCauseRetriever.java
```diff
@@ -46,6 +48,28 @@
         return env;
     }
 
+    public String getCauseUserName(AbstractBuild<?, ?> build) {
```
oleg-nenashev: The method needs to be annotated an documented  properly.
Also the parameter should be <code>Run<?,?></code> unless you need something from the Abstract build

src/main/java/org/jenkinsci/plugins/envinject/service/BuildCauseRetriever.java
```diff
@@ -46,6 +48,28 @@
...
+        List<Cause> buildCauses = causeAction.getCauses();
+        for (Cause cause : buildCauses) {
+            if (isUserCause(cause) || isUserIdCause(cause)) {
+            	return getUserName(cause);
+            }
+        }
+		return null;
+	}
+
+    public String getCauseUserId(AbstractBuild<?, ?> build) {
```
oleg-nenashev: Same

src/main/java/org/jenkinsci/plugins/envinject/service/BuildCauseRetriever.java
```diff
@@ -94,4 +118,39 @@ private static String getTriggerName(Cause cause) {
         return null;
     }
 
+    private static Boolean isUserCause(Cause cause) {
+    	if (cause instanceof UserCause){
```
oleg-nenashev: It is an obsolete cause. Should also check UserIdCause

src/main/java/org/jenkinsci/plugins/envinject/service/BuildCauseRetriever.java
```diff
@@ -94,4 +118,39 @@ private static String getTriggerName(Cause cause) {
...
+    		return true;
+    	}
+    	return false;
+    }
+
+    private static Boolean isUserIdCause(Cause cause) {
```
oleg-nenashev: Both calls should be in a single method IMHO

src/main/java/org/jenkinsci/plugins/envinject/service/BuildCauseRetriever.java
```diff
@@ -1,8 +1,13 @@
 package org.jenkinsci.plugins.envinject.service;
 
+import static com.google.common.base.Joiner.*;
+import static org.apache.commons.lang.StringUtils.*;
```
oleg-nenashev: Please do not use wildcards in the production code imports

oleg-nenashev:
BTW, I still do not see much justification for this change since there is https://wiki.jenkins-ci.org/display/JENKINS/Build+User+Vars+Plugin

zhangzhenyu0819:
sorry, I do oblivious to the build-user-vars-plugin.
Close this PR?

PR #73, opened 2015-09-14
Fix for Unix variable that contain itself in value.

michelepagot:
If one of the variables from file is something like 
GATTO=${GATTO} miao
The plugin iteratively try to solve it. Result is
GATTO=${GATTO} miao miao miao miao miao miao miao miao miao miao miao miao miao miao miao miao miao miao miao miao miao miao miao miao miao miao miao miao miao ....

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectEnvVars.java
+++ b/src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectEnvVars.java
@@ -206,8 +206,7 @@ public class EnvInjectEnvVars implements Serializable {
 
         //Resolve variables against env
         for (Map.Entry<String, String> entry : variables.entrySet()) {
-            if(isSelfRef(entry))
-            {
+            if(isSelfRef(entry)) {
                 entry.setValue(removeSelfRef(entry));
             }
             String value = Util.replaceMacro(entry.getValue(), env);
@@ -285,14 +284,10 @@ public class EnvInjectEnvVars implements Serializable {
             return false;
         }
 
-        if (isSelfRef(env)) {
-            return false;
-    }
-
         return env.getValue().contains("$") && !env.getValue().contains("\\$");
     }
     
-    private boolean isSelfRef(Map.Entry<String, String> env) {
+    private static boolean isSelfRef(Map.Entry<String, String> env) {
 
         if (env.getValue().contains("${".concat(env.getKey()).concat("}"))) {
             return true;
@@ -311,18 +306,6 @@ public class EnvInjectEnvVars implements Serializable {
             return value.replace("\\$", "$");
         }
         return value;
-    }
-
-    public Map<String, String> removeSelfRef(Map<String, String> envVars) {
-        Map<String, String> result = new HashMap<String, String>();
-        for (Map.Entry<String, String> entry : envVars.entrySet()) {
-            if(isSelfRef(entry))
-            {
-                result.put(entry.getKey(), removeSelfRef(entry));
-                logger.info(String.format("Clean self referencing '%s' variable.", entry.getKey()));
-}
-        }
-        return result;
     }
     
     private String removeSelfRef(Map.Entry<String, String> env) {
--- a/src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectEnvVars.java
+++ b/src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectEnvVars.java
@@ -206,7 +206,7 @@ public class EnvInjectEnvVars implements Serializable {
 
         //Resolve variables against env
         for (Map.Entry<String, String> entry : variables.entrySet()) {
-            if(isSelfRef(entry)) {
+            if (isSelfRef(entry)) {
                 entry.setValue(removeSelfRef(entry));
             }
             String value = Util.replaceMacro(entry.getValue(), env);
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectEnvVars.java
```diff
@@ -206,6 +206,10 @@ public void resolveVars(Map<String, String> variables, Map<String, String> env)
 
         //Resolve variables against env
         for (Map.Entry<String, String> entry : variables.entrySet()) {
+            if(isSelfRef(entry))
+            {
```
oleg-nenashev: Please don't add new lines before brackets

src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectEnvVars.java
```diff
@@ -270,18 +274,35 @@ public String resolve(String name) {
...
+        return env.getValue().contains("$") && !env.getValue().contains("\\$");
+    }
+    
+    private boolean isSelfRef(Map.Entry<String, String> env) {
+
+        if (env.getValue().contains("${".concat(env.getKey()).concat("}"))) {
+            return true;
+        }
+        
+        if (env.getValue().contains("$".concat(env.getKey()))) {
```
oleg-nenashev: Won't work correctly with:
```
FOO=something $FOOBAR
```

src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectEnvVars.java
```diff
@@ -270,18 +274,35 @@ public String resolve(String name) {
...
 
-        return value.contains("$") && !value.contains("\\$");
+        if (isSelfRef(env)) {
+            return false;
+    }
+
+        return env.getValue().contains("$") && !env.getValue().contains("\\$");
+    }
+    
+    private boolean isSelfRef(Map.Entry<String, String> env) {
```
oleg-nenashev: It's better to make methods "static" if they don't use the context of the class

src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectEnvVars.java
```diff
@@ -292,4 +313,31 @@ private String removeEscapeDollar(String value) {
         return value;
     }
 
+    public Map<String, String> removeSelfRef(Map<String, String> envVars) {
```
oleg-nenashev: This method seems to be the unused one

src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectEnvVars.java
```diff
@@ -292,4 +313,31 @@ private String removeEscapeDollar(String value) {
...
+                result.put(entry.getKey(), removeSelfRef(entry));
+                logger.info(String.format("Clean self referencing '%s' variable.", entry.getKey()));
+}
+        }
+        return result;
+    }
+    
+    private String removeSelfRef(Map.Entry<String, String> env) {
+        //We replace self variable name unless we are on Windows (Unix only)
+        if ('/' == File.separatorChar) { //unix test
```
oleg-nenashev: I doubt it may work in such way, because File.separatorChar will indicate Jenkins master instead of the target node

oleg-nenashev:
IMO the PR needs some fixes and unit tests.
Please also create a JIRA issue. For me it seems to be a low-priority bug with an obvious workaround (assignment via a transient variable). BTW tthe issue may hit hard for PATH, LD_LIBRARY_PATH, LM_LICENSE_FILE, etc.

michelepagot:
image

image

src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectEnvVars.java
michelepagot: It is the same of removeEscapeDollar

michelepagot:
Please better explain what do you mean with " BTW tthe issue may hit hard for PATH, LD_LIBRARY_PATH, LM_LICENSE_FILE, etc."

src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectEnvVars.java
```diff
@@ -206,6 +206,9 @@ public void resolveVars(Map<String, String> variables, Map<String, String> env)
 
         //Resolve variables against env
         for (Map.Entry<String, String> entry : variables.entrySet()) {
+            if(isSelfRef(entry)) {
```
recena: Please, space after `if`

oleg-nenashev:
I'm keeping :-1: due to https://github.com/jenkinsci/envinject-plugin/pull/73#discussion_r39416970

> Please better explain what do you mean with " BTW tthe issue may hit hard for PATH, LD_LIBRARY_PATH, LM_LICENSE_FILE, etc."

I just figured out use-cases, where your change is useful. BTW, a common approach would be to modify the property file syntax (e.g. "PATH+=/opt/foo/bin") like Jenkins does in its internal APIs

oleg-nenashev:
removed comments from @dummy-lanwen-bot

oleg-nenashev:
#90 has been integrated. IMHO it provides a more elegant fix with some tests, so I think this PR can be closed

PR #77, opened 2015-11-13
[JENKINS-23666] Prefer property files in workspace...

wgreven:
… over property file in current working directory.


oleg-nenashev:
I need to check the docs, because there is a risk of breaking use-cases

oleg-nenashev:
Re-checked it. Yes, it is a breaking change even if it's correct.
I'm find with changing the default resolution order in EnvInject 2.x (to be released soon), but maybe we want to retain the original behavior.

PR #78, opened 2015-11-16
[FIXED JENKINS-31573] New line characters not parsing correctly

Jones2026:
[JENKINS-31573](https://issues.jenkins-ci.org/browse/JENKINS-31573)

The main reason I created this pull request was because I was unable to have a properties file that parsed \n characters correctly.

To do this I eliminated the use of the custom properties class and instead just used a standard Java properties object (http://docs.oracle.com/javase/6/docs/api/java/util/Properties.html) to store/parse the properties content and/or properties file. This ensures we stick with the standard way Java handles properties files.

Also updated the tests to treat \ as a line continuation and not attempt to parse that as \n within the properties file. This mirrors how Java treats properties files and allows us to use both \ and \n when desired and get consistent results. 

Lastly, removed some code duplication in the PropertiesLoader class by extracting a method that both properties file and properties content use. This also helps ensure that both behave in the same fashion.


recena:
@Jones2026 It would be nice to create a JIRA issue to track this.

Jones2026:
Completely forgot to do that, thanks! Here is the JIRA issue:
https://issues.jenkins-ci.org/browse/JENKINS-31573

sassda:
Was this maybe fixed somewhere else? I still have this issue on my side

Jones2026:
Definitely not fixed that I know of. This pull request just needs to be merged to fix the issue.

duncanmak:
@Jones2026 This is great! I just came across this and was pulling my hair out reading the code in `SortedProperties.java` - I was contemplating writing a patch myself, but I see that you've done a better job at incising the monstrosity than I could have done myself.

Thanks again!

👍  Let's get this merged in.

Jones2026:
@oleg-nenashev Can you merge this by chance or do you know who can?

src/main/java/org/jenkinsci/plugins/envinject/util/SortedProperties.java
```diff
@@ -1,273 +0,0 @@
...
-
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.Reader;
-import java.util.LinkedHashMap;
-
-/**
- * @author Gregory Boissinot
- */
-public class SortedProperties extends LinkedHashMap<Object, Object> {
```
oleg-nenashev: Deletion of this class formally breaks binary compatibility

oleg-nenashev: It would be better to just deprecate it

Jones2026: I added the class back in and marked it as deprecated.

oleg-nenashev:
Need to review it again

Jones2026:
Any other comments or concerns on this?

spa-87:
Hi, it would be great if that issue was fixed asap :)

mputilin:
almost a year passed. any plans to merge/fix that?

Jones2026:
@oleg-nenashev Anything else I can do to help get this merged?

oleg-nenashev:
After some testing the change looks reasonable to me.
Sorry for the long delay. Integrating

PR #79, opened 2015-11-27
Fix an NPE when sensitiveVariables is a null object

castorls:
I've encountered a NullPointerException on my environment and this is the fix I've done.

Regards,


oleg-nenashev:
I suppose it has been already fixed

oleg-nenashev:
Actually this misuse case has not been fixed. Somebody needs to rework the PR.
Would be great to know how you get into this issue, but it's not mandatory

oleg-nenashev:
Checked all open-source plugins and the core. 
getSensitiveBuildVariables() cannot return null in all known implementations. 

@castorls If you have a stacktrace, please respond. Closing the PR by now

PR #80, opened 2015-12-01
[JENKINS-31829] - Add currentListener Groovy variable

Fiouz:


src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectEnvVars.java
```diff
@@ -140,6 +140,7 @@ public int executeScript(boolean loadFromMaster,
             groovyShell.setVariable(entryVariable.getKey(), entryVariable.getValue());
         }
         groovyShell.setVariable("out", logger.getListener().getLogger());
+        groovyShell.setVariable("currentListener", logger.getListener());
```
oleg-nenashev: Other Groovy steps use the "listener" variable, but I agree that this PR should follow the legacy approach

oleg-nenashev:
Would be nice to have a Unit test for it

oleg-nenashev:
Merge conflict

PR #87, opened 2016-02-20
[FIXED JENKINS-30028] Remove logging from password wrapper decorateLogger

StevenGBrown:
[JENKINS-30028](https://issues.jenkins-ci.org/browse/JENKINS-30028)

When the Timestamper build wrapper runs first, the messages logged produced by EnvInjectPasswordWrapper.decorateLogger do not receive timestamps. Fix this by logging these messages from setUp instead.

I was able to reproduce this by using the [ordinal value](https://github.com/jenkinsci/jenkins/blob/2893fd3/core/src/main/java/hudson/Extension.java#L82) to force the Timestamper plugin to load first.

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/envinject-plugin/87)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPasswordWrapper.java
+++ b/src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPasswordWrapper.java
@@ -79,6 +79,14 @@ public class EnvInjectPasswordWrapper extends BuildWrapper {
 
     @Override
     public Environment setUp(AbstractBuild build, Launcher launcher, BuildListener listener) throws IOException, InterruptedException {
+        EnvInjectLogger logger = new EnvInjectLogger(listener);
+        if (isInjectGlobalPasswords()) {
+            logger.info("Inject global passwords.");
+        }
+        if (isMaskPasswordParameters()) {
+            logger.info("Mask passwords passed as build parameters.");
+        }
+
         return new Environment() {
         };
     }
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPasswordWrapper.java
```diff
@@ -79,6 +79,14 @@ public void setPasswordEntries(EnvInjectPasswordEntry[] passwordEntries) {
 
     @Override
     public Environment setUp(AbstractBuild build, Launcher launcher, BuildListener listener) throws IOException, InterruptedException {
+        EnvInjectLogger logger = new EnvInjectLogger(listener);
+        if (isInjectGlobalPasswords()) {
+            logger.info("Inject global passwords.");
```
oleg-nenashev: This message makes no sense now since it does not identify the stage in DecorateLogger()

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPasswordWrapper.java
```diff
@@ -79,6 +79,14 @@ public void setPasswordEntries(EnvInjectPasswordEntry[] passwordEntries) {
...
+        }
+        if (isMaskPasswordParameters()) {
+            logger.info("Mask passwords passed as build parameters.");
```
oleg-nenashev: "will be passed"?

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPasswordWrapper.java
StevenGBrown: What would you like me to do? Shall I remove it?

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPasswordWrapper.java
StevenGBrown: OK. Will change to "Mask passwords that will be passed as build parameters."

StevenGBrown:
Merged. I'm happy to follow up with another pull request about changing the log messages if you would like.

oleg-nenashev:
@StevenGBrown BTW you have not addressed comments like this one: https://github.com/jenkinsci/envinject-plugin/pull/87#discussion_r72615250
I'm not against merging it since I've dropped a ball on review

StevenGBrown:
Great. I've also pushed that change to the log message that you wanted: 1955123

PR #90, opened 2016-05-16
JENKINS-30467 support common cases of recursive variables

XN137:
on a matrix job we had something like this line in "Properties Content":
`PATH=$WORKSPACE/git_checkout/bin:${PATH}`

it expanded recursively (`WORKSPACE` being quite long already for matrix runners) and `PATH` grew very large and I think it caused build failures later, where calling bash failed with:
`Argument list too long`

duplicate issues: JENKINS-30446, JENKINS-30445, JENKINS-30444

similar PR: #73 

Regarding changes in `PropertiesVariablesRetriever`:
this logic should probably be re-used more often across the whole code base instead of plain `putAll`, which just overwrites values.
Currently there are still many inconsistencies regarding referencing previously defined variables (see comment in Test `injectTextExtendSysEnvVar` for an example) but fixing all those issues is outside of the scope of this PR.

I adjusted `PropertiesGetterTest` because it failed on my system (ubuntu xenial, jdk 1.8.0_91) with inverse order:
```
Failed tests: 
  PropertiesGetterTest.getPropertiesContentThreeElements:48 expected:<key[3=value3
key2=value2
key1=value1]> but was:<key[1=value1
key2=value2
key3=value3]>
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectEnvVars.java
+++ b/src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectEnvVars.java
@@ -201,7 +201,6 @@ public class EnvInjectEnvVars implements Serializable {
         return variables;
     }
 
-
     public void resolveVars(Map<String, String> variables, Map<String, String> env) {
 
         //Resolve variables against env
@@ -218,7 +217,11 @@ public class EnvInjectEnvVars implements Serializable {
             int previousNbUnresolvedVar = nbUnresolvedVar;
             nbUnresolvedVar = 0;
             for (Map.Entry<String, String> entry : variables.entrySet()) {
-                String value = Util.replaceMacro(entry.getValue(), variables);
+                Map<String, String> resolveVariables = new HashMap<String, String>(variables);
+                // avoid self reference expansion
+                // FOO=$FOO-X -> FOO=$FOO-X-X -> FOO=$FOO-X-X-X etc.
+                resolveVariables.remove(entry.getKey());
+                String value = Util.replaceMacro(entry.getValue(), resolveVariables);
                 entry.setValue(value);
                 if (isUnresolvedVar(value)) {
                     nbUnresolvedVar++;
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectEnvVars.java
```diff
@@ -218,9 +217,16 @@ public void resolveVars(Map<String, String> variables, Map<String, String> env)
...
+                String key = entry.getKey();
+                String value = entry.getValue();
+                Map<String, String> resolveVariables = new HashMap<String, String>(variables);
+                // avoid self reference expansion
+                // FOO=$FOO-X -> FOO=$FOO-X-X -> FOO=$FOO-X-X-X etc.
+                resolveVariables.remove(key);
+                String newValue = Util.replaceMacro(value, resolveVariables);
+                entry.setValue(newValue);
+                boolean variableChanged = !newValue.equals(value);
+                if (variableChanged && isUnresolvedVar(newValue)) {
```
oleg-nenashev: Are you sure that <code> variableChanged</code> is required here. For me it seems the counter won't be incremented properly if there is unresolved variables

oleg-nenashev:
Sorry for the delay. My action item is to review it

src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectEnvVars.java
XN137: i think it is not required, but i felt it is more correct.
In the loop we resolve variables against themselves... if a variable's value did not change by resolving and it is still not resolved yet, it wont resolve in any of the future iterations, thus we should not consider it unresolved with regard to the continuation of the loop.

In practice it probably just saves a single loop at the end.
So if you prefer to keep the previous logic, i can change it.

XN137: @oleg-nenashev so do you want me to remove `variableChanged` ?

oleg-nenashev: Yes, it would be better to remove it (or to implement a test case for the potential issue I've mentioned above)

XN137:
@oleg-nenashev I've made the changes you asked for, sorry for the delay.

oleg-nenashev:
I'll merge it manually

oleg-nenashev:
The fix has been integrated towards 1.93

XN137:
thanks for merging.

some things i noticed:
- JENKINS-30446 was also a duplicate of this issu and not resolved yet in JIRA
- https://wiki.jenkins-ci.org/display/JENKINS/EnvInject+Plugin changelog does not list any of the resolved tickets (probably should be JENKINS-30467)

PR #97, opened 2016-07-08
Fix JENKINS-36545

nfalco79:
Move initialization of WORKSPACE variable before run groovy script.
Some code clean-up.
Fix bug that eat exception message in case of script errors.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/envinject/EnvInjectBuildWrapper.java
+++ b/src/main/java/org/jenkinsci/plugins/envinject/EnvInjectBuildWrapper.java
@@ -3,6 +3,7 @@ package org.jenkinsci.plugins.envinject;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.Serializable;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
@@ -18,6 +19,7 @@ import hudson.scm.SCM;
 import hudson.tasks.BuildWrapper;
 import hudson.tasks.BuildWrapperDescriptor;
 import net.sf.json.JSONObject;
+
 import org.jenkinsci.lib.envinject.EnvInjectLogger;
 import org.jenkinsci.plugins.envinject.service.EnvInjectActionSetter;
 import org.jenkinsci.plugins.envinject.service.EnvInjectEnvVars;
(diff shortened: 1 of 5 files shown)


src/main/java/org/jenkinsci/plugins/envinject/EnvInjectBuildWrapper.java
```diff
@@ -104,8 +101,8 @@ public void buildEnvVars(Map<String, String> env) {
                     env.putAll(resultVariables);
                 }
             };
-        } catch (Throwable throwable) {
-            logger.error("Problems occurs on injecting env vars as a build wrap: " + throwable.getCause());
+        } catch (Exception e) {
```
oleg-nenashev: I'm not sure this is a right change. Groovy engine may case many issues like OOM, and this change causes unrecoverable build failure

oleg-nenashev:
Generally looks good to me even if there are unrelated changes

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectBuildWrapper.java
nfalco79: previous in case of OOM the build was set to FAILURE and there is no chance to change the state of build.

Typycally catch Throwable is bed practice and the OOM happens in the same JVM of jenkins so, now IT would be informed that jenkins instance should be restarted or some JVM setting be should be adjust to avoid again issue.
Previously an OOM (for example) was eated and an incomprensible message was printed (null :\) so anyone would be alern of the problem until happen in other point.

nfalco79:
Except Throwable changes other one it's only a fix of sonar violation.

nfalco79:
Without the fix to exception block is almost impossible understand where is the error in the groovy script
For example try this portion of code in the actual plugin version:
```
def now = new Date(System.currentTimeMillis())
def firstDay = new GregorianCalendar(2013, 10, 11).time
def millisInWeek = 1000 * 60 * 60 * 24 * 7
def diff = (now.getTime() - firstDay.getTime()) / millisInWeek
def sprint = (long) Math.ceil(diff / 3)

def qualifier = String.format("%02d%06d", sprint, Integer.parseInt(BUILD_NUMBER))

def pomContent = new File(WORKSPACE + '/pom.xml').getText('UTF-8')
def pom = new XmlSlurper().parseText(pomContent)
def version = pom.version.text()
if (version.contains("SNAPSHOT") || version.contains("RELEASE")) {
    version = version.replaceAll("-(SNAPSHOT|RELEASE)", '')
    version = String.format("%s.v%s", version, qualifier)
}

def map = [VERSION: version]
return map
```

nfalco79:
Back to Throwable to facilitate merge and discussion.
Add test case for the WORKSPACE fix
Add test case to exception eating (now EnvInjectBuildWrapper threats exception just like EnvInjectBuilder does).

Re enable some test classes not collected by JUnit becase its name does not end with Test keyword. Fix one of them that does not work corect if run under Jenkins job due enviroment variables.

oleg-nenashev:
Should be fine. 👍

nfalco79:
Any news?

nfalco79:
I change the first commit message to [FIXED JENKINS...] as suggested [here](https://wiki.jenkins-ci.org/display/JENKINS/Pull+Request+to+Repositories) to close the JIRA ticket by the bot.

oleg-nenashev:
Fighting with the parent POM right now

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectConstants.java
```diff
@@ -0,0 +1,16 @@
+package org.jenkinsci.plugins.envinject;
+
+/**
+ * Define the environment variables
+ */
+public final class EnvInjectConstants {
```
oleg-nenashev: I think this class should be made restricted, because the API may need some refactoring. <code>@Restricted(NoExternalUse.class)</code>. I can do it after the merge

oleg-nenashev:
I'll fix the Restricted case later

PR #100, opened 2016-09-18
Update core dependency to 1.609.3

oleg-nenashev:
It supersedes #82. The reason for the upgrade is a need to pick security fixes available in newer Jenkins core versions. 1.609.3 is the target release, because newer version do not support Java 6.

After this release I hope to jump to Java 7 shortly since it could help with plugin code quality fixes.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.12</version>
+        <version>2.15</version>
     </parent>
 
     <artifactId>envinject</artifactId>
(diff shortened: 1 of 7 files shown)


oleg-nenashev:
@reviewbybees

pom.xml
```diff
@@ -60,8 +60,8 @@
         <ivy.plugin.version>1.21</ivy.plugin.version>
         <mockito.version>1.10.19</mockito.version>
 
-        <jenkins.version>1.532.3</jenkins.version>
-        <jenkins-test-harness.version>1.532.3</jenkins-test-harness.version>
+        <jenkins.version>1.609.3</jenkins.version>
+        <jenkins-test-harness.version>1.609.3</jenkins-test-harness.version>
```
jglick: Delete this override and pick up the floating 2.x harness if you can.

pom.xml
```diff
@@ -115,8 +115,8 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>matrix-project</artifactId>
-            <version>1.4</version>
-            <scope>test</scope>
+            <version>1.7</version>
+            <type>jar</type>
```
jglick: `type=jar` is the default, need not specify it. Are you intentionally making this into a plugin dependency? I guess because it already was, but implicitly prior to split?

amuniz:
:bee:

pom.xml
oleg-nenashev: Yes, it was before the split. We were using newer Matrix project just to fix integration bug, but the dependency was already there

pom.xml
oleg-nenashev: Maybe next time, there some issues with the test cases

PR #102, opened 2016-09-23
Simplify form posting

rsandell:
Remove legacy json handling and replace with more formal DataBound handling.
Needed for a future fix.

@reviewbybees esp @oleg-nenashev


src/main/java/org/jenkinsci/plugins/envinject/EnvInjectJobProperty.java
```diff
@@ -34,6 +36,17 @@
 
     private transient EnvInjectJobPropertyContributor[] contributorsComputed;
 
+    @DataBoundConstructor
+    public EnvInjectJobProperty(EnvInjectJobPropertyInfo info) {
+        this.info = info;
+    }
+
+    @Deprecated
+    public EnvInjectJobProperty() {
```
oleg-nenashev: Also <code>@Restricted</code>

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectJobProperty.java
```diff
@@ -163,45 +181,12 @@ public String getHelpFile() {
...
-                    envInjectJobProperty.setKeepBuildVariables(onJSONObject.getBoolean("keepBuildVariables"));
-                    envInjectJobProperty.setOverrideBuildParameters(onJSONObject.getBoolean("overrideBuildParameters"));
-
-                    //Process contributions
-                    setContributors(req, envInjectJobProperty, onJSONObject);
-
-                    return envInjectJobProperty;
-                }
+            if (formData.optBoolean("on")) {
+                return (EnvInjectJobProperty)super.newInstance(req, formData);
```
oleg-nenashev: NIT: Maybe newInstance() handler implementation is not required at all

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectJobPropertyInfo.java
```diff
@@ -41,4 +45,17 @@ public String getGroovyScriptContent() {
...
+    @Override
+    public Descriptor<EnvInjectJobPropertyInfo> getDescriptor() {
+        return Jenkins.getInstance().getDescriptorByType(DescriptorImpl.class);
+    }
+
+    @Extension
+    public static class DescriptorImpl extends Descriptor<EnvInjectJobPropertyInfo> {
+        @Override
+        public String getDisplayName() {
+            return "EnvInjectJobPropertyInfo";
```
oleg-nenashev: This displayName is never used, hence fine

src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectJobPropertyInfo/config.jelly
```diff
@@ -0,0 +1,41 @@
...
+                <f:checkbox />
+            </f:entry>
+        </j:when>
+        <j:otherwise>
+            <f:hidden field="loadFilesFromMaster"
+                      value="false"/>
+        </j:otherwise>
+    </j:choose>
+
+</j:jelly>
```
oleg-nenashev: NIT: line break

src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectJobProperty/help-propertiesContent.html
```diff
@@ -1,16 +0,0 @@
-<div>
-    <p>Give a set of key/value (one variable per line): <code>KEY=VALUE</code>.</p>
```
oleg-nenashev: So thiese help files have been actually moved from what I see...

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectJobProperty.java
rsandell: I copied the code from OptionalJobProperty that isn't available in the current core dep.
I was first trying to remove the on field completely and just check instance, but there are some logic in the backend using it, so I let it be.

src/main/resources/org/jenkinsci/plugins/envinject/EnvInjectJobProperty/help-propertiesContent.html
rsandell: Before there where duplicates in both the JobProperty and the BuildWrapper, since moving the config into the Info class some duplicates could be removed.

rsandell:
My biggest concern with the change was the contributors field. I have manually tested those with `ontrack-plugin` and 'shared-objects-plugin' and they work.
I haven't changed the data-model so reading old data should be fine. But I have some unit tests lined up for that for my next fix.

oleg-nenashev:
Help in Global configuration in /configure does not work anymore. Maybe not a regression in this PR. The variable injection from it does not work as well, need to troubleshoot.

screen shot 2016-09-23 at 15 57 40

rsandell:
That is strange, I tested that. Will take another look

oleg-nenashev:
Variable injection actually works fine. It just needs a restart of the master according to the design. Weird design :)

rsandell:
It works on my machine, but for me it also says `Prepare an environment for the run` or `Inject environment variables to the build process`, so I'm not sure where you did your screenshot from.

oleg-nenashev:
@rsandell This screenshot is from "/configure", not from "/job/XXX/configure". It comes from EnvInjectNodeProperty

🐛 by now

rsandell:
@oleg-nenashev good catch I haven't checked the global config at all.

oleg-nenashev:
I'll fix that help message on my own. It still needs some details compared to the original version

PR #103, opened 2016-09-23
Total Annotation - use FindBugs to cleanup the mess

oleg-nenashev:
During the review of #102 discovered a potential NPE. And then another one...
So I want to just cover the entire code by annotations in order to cleanup such pitfalls

CC @rsandell, who may have some merge fun with his ongoing changes. We need to determine what goes first


oleg-nenashev:
@jenkinsci/code-reviewers @reviewbybees Just in case anybody wants to review it

oleg-nenashev:
Merging as an owner. If somebody wants to provide feedback, please free to follow-up. I'm planning a new release tomorrow.

PR #115, opened 2017-04-01
fix groovy script example in help text

saadmufti:
In the example in the help text, the key in the returned map should be a string from what I have tested. Otherwise groovy throws an exception like the following:

groovy.lang.MissingPropertyException: No such property: COMPUTE_VAR for class: xxx


oleg-nenashev:
Sorry, the latest security changes caused a merge conflict. Could you merge your change with the master?

saadmufti:
Hi, I rebased against the latest master.

saadmufti:
Sorry e-mail and author name were wrong in latest commit, I changed those too

oleg-nenashev:
Thanks!

PR #120, opened 2017-05-29
[JENKINS-32428] Escape xml and json outputs in REST API

Rechi:
If environment variables or their values contain special characters, there is no valid xml respectively json output.


Rechi:
I haven't searched for a ticket and haven't created one.

oleg-nenashev:
https://issues.jenkins-ci.org/browse/JENKINS-32428 works for me

oleg-nenashev:
Thanks!

PR #121, opened 2017-06-19
[FIXED JENKINS-44263] - Get rid of explict cached value access.

oleg-nenashev:
In EnvInject 3.1 I was doing refactoring for better Pipeline Support, and I have somehow removed cache calculation in the Build lazy loading chain. It caused regression (missing data) in `EnvInjectPluginAction#getEnvInjectVarList()`, which was always reading the cache value instead of invoking the Getter method. And there was no roundtrip tests...

https://issues.jenkins-ci.org/browse/JENKINS-44263

@reviewbybees

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/envinject/EnvInjectVarListTest.java
+++ b/src/test/java/org/jenkinsci/plugins/envinject/EnvInjectVarListTest.java
@@ -1,0 +1,61 @@
+package org.jenkinsci.plugins.envinject;
+
+import hudson.model.FreeStyleBuild;
+import hudson.model.FreeStyleProject;
+import hudson.model.Run;
+import java.util.Map;
+import javax.annotation.Nonnull;
+import org.junit.Assert;
+import static org.junit.Assert.assertEquals;
+import org.junit.Rule;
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/envinject/EnvInjectVarListTest.java
```diff
@@ -0,0 +1,61 @@
...
+        
+        // Reload and check vars
+        // build.reload() does not assign parents for RunAction2, hence we apply a workaround
+        p.doReload();
+        final Run<?, ?> reloadedBuild = p.getBuildByNumber(build.getNumber());
+        {
+            EnvInjectVarList varList = getVarListOrFail(reloadedBuild);
+            Map<String, String> envMap = varList.getEnvMap();
+            Assert.assertNotNull("EnvInject vars list is null for the run after the reload", envMap);
+            Assert.assertTrue("TEXT_VAR is not present in the list", envMap.containsKey("TEXT_VAR"));
```
jglick, line 48: Catches regression fixed by https://github.com/jenkinsci/envinject-lib/pull/19.

PR #123, opened 2017-06-28
[JENKINS-32428] Escape xml and json outputs in REST API - fix 2 (#120)

Rechi:
I added the replaces in wrong order at #120.
`"C:\Program Files\Java\jre1.8.0_131\bin"` got escaped with `\\"C:\\Program Files\\Java\\jre1.8.0_131\\bin\\"` which isn't valid json. The correct escaped string is `\"C:\\Program Files\\Java\\jre1.8.0_131\\bin\"`

@oleg-nenashev ping


oleg-nenashev:
Ack, will spin the release soon. That's why having autotests is good :(

PR #125, opened 2017-09-29
[JENKINS-26583] - Do not suppress environment variables contributed by extensions

oleg-nenashev:
It is a partial version of #124, which does not fix all tests created by @olivergondza. But it at least resolves the issue reported in [JENKINS-26583](https://issues.jenkins-ci.org/browse/JENKINS-26583), at least on the common paths.

- [x] - Integrated #37 with master, polished tests a bit
- [x] - Prevented `EnvInjectPluginAction` from overriding variables on late stages, 3 tests fixed
- [x] - Change the priority of `EnvInjectListener` to call it first and to let others override the EnvVars
- [x] - `EnvInjectPluginAction` still overrides build parameters, just because it is covered by tests

More fixes are to come later.

@reviewbybees @olivergondza @recena


carltonbrown:
I verified that this PR fixes https://issues.jenkins-ci.org/browse/JENKINS-26583 .   Please merge asap!

oleg-nenashev:
Retriggering CI

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginAction.java
```diff
@@ -76,11 +101,82 @@ public String transformEntry(String key, String value) {
...
+        }
+
+        // Other extension points may contribute other variable values
+        // before contributing actions is invoked. See AbstractBuild#getEnvironment()
+        // We take the externally updated variables as a source of truth and just override the missing ones
+        Map<String, String> overrides = null;
+        for (Map.Entry<String, String> storedVar : currentEnvMap.entrySet()) {
+            final String varName = storedVar.getKey();
+            final String storedValue = storedVar.getValue();
+            final String envValue = env.get(storedVar.getKey());
```
nfalco79: I suggest to use local variable varName defined 2 row above

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginAction.java
```diff
@@ -76,11 +101,82 @@ public String transformEntry(String key, String value) {
...
+                    if (envValue.equals(parameterValue)) { // defined by parameter and not already overridden
+                        LOGGER.log(Level.CONFIG, "Build {0}: Overriding value of {1} defined by the parameter value. New value is {2}, was {3}",
+                                new Object[] {build, varName, storedValue, envValue});
+                        env.put(varName, storedValue);
+                        usedExternalValue = false;
+                    }
+                }
+
+                if (usedExternalValue) { // The value was overridden, let's update the cache
+                    LOGGER.log(Level.CONFIG, "Build {0}: Variable {1} is defined externally, overriding the stored value {2} by {3}",
```
nfalco79: Are password logged in plain text?

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginAction.java
```diff
@@ -76,11 +101,82 @@ public String transformEntry(String key, String value) {
...
+                        new Object[] {build, varName, storedValue});
+                env.put(varName, storedValue);
+            } else if (!envValue.equals(storedValue)) {
+                // If the value is defined by the Parameters, we actually override them
+                // See org.jenkinsci.plugins.envinject.EnvInjectJobPropertyTest#shouldOverrideBuildParametersIfEnabled()
+                final EnvVars parameterEnvVars = getParameterEnvVars();
+                boolean usedExternalValue = true;
+                if (parameterEnvVars != null) {
+                    String parameterValue = parameterEnvVars.get(varName);
+                    if (envValue.equals(parameterValue)) { // defined by parameter and not already overridden
```
nfalco79: In the previous comment you refer to `shouldOverrideBuildParametersIfEnabled()` settings.
I do not see where the value is taken, I would expect that value in && here

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginAction.java
oleg-nenashev: Agreed

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginAction.java
oleg-nenashev: shouldOverrideBuildParametersIfEnabled() is a test, there is no such configuration option

oleg-nenashev:
@nfalco79 Thanks for catching it! Updated the PR

nfalco79:
Also for me seems work (PATH env var is valid)

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginAction.java
nfalco79: I remember that option in the job configure page:
image

oleg-nenashev:
I still anticipate some regressions, likely I will add a compatibility warning. In #124 I tried to do more things about the resolution order (e.g. Extension listing order), but it is a subject for the next pull requests.

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginAction.java
oleg-nenashev: agreed, needs investigation

dwnusbaum: I agree with @nfalco79 here, I think since the existing behavior is being changed anyway, it makes sense to start checking that option here. From what I can see, that option is currently only used in the [EnvInjectBuildVariableContributor](https://github.com/jenkinsci/envinject-plugin/blob/127850b92b01a0430b1edc7ce428e7d55626af6b/src/main/java/org/jenkinsci/plugins/envinject/EnvInjectBuildVariableContributor.java#L34) class which does something similar to this piece of code. Tests pass for me locally with the change:
```
EnvInjectJobProperty envInjectJobProperty = RunHelper.getEnvInjectJobProperty(build);
if (parameterEnvVars != null && envInjectJobProperty != null && envInjectJobProperty.isOverrideBuildParameters() && parameterEnvVars != null) {
    String parameterValue = parameterEnvVars.get(varName);            
```

oleg-nenashev:
@dwnusbaum @nfalco79 done

oleg-nenashev:
My plan is to release it today/tomorrow together with https://github.com/jenkinsci/envinject-lib/pull/11 from @dwnusbaum

oleg-nenashev:
@reviewbybees done

oleg-nenashev:
Merging without squash since there are multiple contributors. No plans to backport this tricks anyway

oleg-nenashev:
The fix caused regressions (no wonder), I am blacklisting the release for a while to investigate them: https://github.com/jenkins-infra/backend-update-center2/pull/167 . Likely I will respin the release and disable the new behavior by default (and add a flag in UI)

dwnusbaum:
@oleg-nenashev I reproduced the reported issues and agree with blacklisting it for now. Hiding the new functionality behind a checkbox that says something like "Do not override environment variables defined elsewhere" makes sense to me, and I think we should also add a few tests to prevent a regression in the future.

nfalco79:
> Hiding the new functionality behind a checkbox that says something like "Do not override environment variables defined elsewhere"

Checkbox solution will not work because the issue JENKINS-26583 happens just installing this plugin.
With that kind of solution means that I have to tick the option in every job / pipeline just because the plugin is in the system.

Actually I had install 2.1.4 in our system and seems the PATH variable is ok because the nodejs command is available. Anyway let me do other try

nfalco79:
Just to be informed, what is the regression?

dwnusbaum:
@nfalco79 [JENKINS-47364](https://issues.jenkins-ci.org/browse/JENKINS-47364) and [JENKINS-47370](https://issues.jenkins-ci.org/browse/JENKINS-47370), users relied on environment variables defined elsewhere being overridden by the plugin, which is exactly the functionality that this PR removed.

nfalco79:
JENKINS-47364 seems a real regression where variables are not substituted.

Instead in JENKINS-47370 if I interpreted right the option "Override Build Parameters" he miss to enable it.

dwnusbaum:
For JENKINS-47370 I tested with that setting but it was still broken. From what I can tell the user just has it a set as a environment variable in Jenkins' global properties configuration, and the "Override Build Parameters" setting only deals with parameters, not global environment variables.

oleg-nenashev:
@nfalco79 In any case, I am not going to revert the fix.

My plan is to temporary disable it while I am figuring out what to do with regressions in non-Pipeline use-cases. In the case of Pipeline, we can break whatever we want in the current state (there is a warning in Wiki). Feature flag is something under consideration. If the impact is not high, I also consider keeping the new behavior by default.

PR #127, opened 2017-10-13
Add compatibility notice for JENKINS-47364 and JENKINS-47370

oleg-nenashev:
Instead of blacklisting of the plugin now, I decided to add a compatibility notice for now while I investigate the issue and create test cases for the reported regressions. [Wiki](https://wiki.jenkins.io/display/JENKINS/EnvInject+Plugin) is explicit enough of the notice reasons.

@nfalco79 @dwnusbaum

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -78,9 +78,10 @@
               Incompatible changes:
                 1.93 - Fix of JENKINS-31573 changed the property file parsing approach. 
                      Now it follows the Java property file standard.
-                2.0: - Security fixes (SECURITY-256) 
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
Also @reviewbybees

oleg-nenashev:
After that https://github.com/jenkins-infra/backend-update-center2/pull/167 can be closed

oleg-nenashev:
@reviewbybees done. Will ship it as is, it does not break anything compared to 2.1.4

PR #128, opened 2017-10-18
Adding stop build support for env-inject groovy script

akomakom:
The need to abort the build in the groovy script is something I stumble upon a lot.  It's easy enough (throw), but results in a build failure.  I'd like to disambiguate a truly failed (red) build from a "I decided not to run this build" by changing its build status.

Note: I couldn't get the test to work exactly like real Jenkins - the test interrupts the job but it runs anyway, while in real life the job does not continue.  I don't know why that is.

Let me know if there are better ways to do all this...

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/envinject/EnvInjectListener.java
+++ b/src/main/java/org/jenkinsci/plugins/envinject/EnvInjectListener.java
@@ -53,6 +53,10 @@ public class EnvInjectListener extends RunListener<Run> implements Serializable 
             } catch (Run.RunnerAbortedException rre) {
                 logger.info("Fail the build.");
                 throw new Run.RunnerAbortedException();
+            } catch (EnvInjectStopBuildException stopException) {
+                logger.info("Stopping build with result " + stopException.getResult() + ": " + stopException.getLocalizedMessage());
+                build.getExecutor().interrupt(stopException.getResult());
+                return null;
             } catch (EnvInjectException e) {
                 e.printStackTrace(listener.error("SEVERE ERROR occurs"));
                 throw new Run.RunnerAbortedException();
--- a/src/main/java/org/jenkinsci/plugins/envinject/EnvInjectStopBuildException.java
+++ b/src/main/java/org/jenkinsci/plugins/envinject/EnvInjectStopBuildException.java
@@ -1,0 +1,24 @@
+package org.jenkinsci.plugins.envinject;
+
+import hudson.model.Result;
+import org.jenkinsci.lib.envinject.EnvInjectException;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Signals the plugin to stop the build and set the build result as specified,
+ * as opposed to Failure in the case of a RuntimeException
+ */
+public class EnvInjectStopBuildException extends EnvInjectException {
+
+    private final Result result;
+
+    public EnvInjectStopBuildException(@Nonnull String s, @Nonnull Result result) {
+        super(s);
+        this.result = result;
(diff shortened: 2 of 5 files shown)


akomakom:
Typical use case (all in env-inject groovy script)

1. Get and inject something from external resources. Examples of my actual use cases:
   * Test count from another job (TestResultAction.getTotalCount() )
   * Metadata from an external build repository
   * SVN revisions built by some other job
   * Check the time (workaround for the [Stash PR cron fiasco](https://support.cloudbees.com/hc/en-us/articles/115003358408)) and fail if the time is wrong.  Yes I am seriously running a job every hour and checking if hour == 8.  It's sad.
   * Checking if a business condition permits the job to run (some API or whatnot)
   * Biweekly cron.  Another sad case, there is no way to implement "run job every 2 weeks" in cron, so I run it weekly and abort on odd weeks.  Again, I don't want any red builds.
2. If the external resources are unavailable (eg not yet ready, job hasn't run, not a good time, etc), I want to abort the build.  I'd like this to look (at a glance) different from a general build failure.  The logical status to set the result to would be ABORTED, but I see no reason why someone may not want UNSTABLE.

The reason why I think that this logic fits EnvInject is that the groovy script is the piece of code that knows whether the preconditions (external resources) permit the job to run or not.   The plugin fills a unique niche - it is the one thing that allows code to run before the build.  

The alternatives are ugly:

1. Wrap the whole build in a Conditional Build Step and check some injected variable.
2. Add another groovy step that attempts to manipulate the build status and fail (again checking injected variables, splitting the logic).  I've found that stopping the build and setting the status arbitrarily is not trivial.
3. Split the job into two...

I do admit that a brand new "Conditional build" plugin may be cleaner (ideally the plugin would simply prevent execution altogether rather than marking build as ABORTED/etc).  However, in my use cases the functionality seems to correlate: it's always "inject or do not run at all", so even if such a plugin existed, there would still be extra configuration in my case.

akomakom:
I rebased to remove ShellStep since cloudbees seems to test this PR on windows

akomakom:
Also, I considered providing some sort of a closure to reduce the verbosity of the stop build mechanism (eg listener.stopBuild(ABORTED) instead of a throw) - but I didn't think that I would do it optimally, so if you have any thoughts on this, let me know.

oleg-nenashev:
I will try to review it once I cleanup the regressions reported in 2.1.4

oleg-nenashev:
After some consideration I still do not feel it is a good idea to add it to the plugin directly.

@jenkinsci/code-reviewers @jglick @abayer @svanoort What do you think about adding a more generic handling logic to the Script Security component? E.g. an equivalent of the `AbortException` in the core. Or it could be done via a new API plugin.

akomakom:
@oleg-nenashev I am not exactly sure that I understand your proposal.  Are you recommending core support for some sort of AbortException that can be thrown from anywhere?  That seems to be roughly equivalent to this PR, so, sure.

oleg-nenashev:
@akomakom Yes, I propose to add such generic exception type to https://github.com/jenkinsci/script-security-plugin/ or even to https://github.com/jenkinsci/groovy-sandbox.

I still do not think it's reasonable to do it only in this plugin

akomakom:
I agree, @oleg-nenashev, besides, I can see this being useful in other groovy plugins.  I'm not sure if I will attempt this myself though.

oleg-nenashev:
@akomakom Thanks for the contribution anyway! Please do not hesitate to create a JIRA ticket to Script security and to refer this pull request

PR #129, opened 2018-01-04
PCT updates

jglick:
POM updates plus failures I noticed running against newer cores.

@reviewbybees esp. @oleg-nenashev


jglick:
@reviewbybees done

PR #132, opened 2018-09-21
Use safe reflection for Java 11

jeffpearce:
This PR removes unsafe reflection to remove warnings on the build agent when the plugin is used.

Please note that it's dependent upon first merging https://github.com/jenkinsci/jenkins/pull/3639, and the version number in this PR needs to be updated once the version is known (see TODO comment in the code)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectMasterEnvVarsSetter.java
+++ b/src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectMasterEnvVarsSetter.java
@@ -45,6 +45,8 @@ public class EnvInjectMasterEnvVarsSetter extends MasterToSlaveCallable<Void, En
                 enVars.remove("MAVEN_OPTS");
             }
             
+            // TODO: this is not thread safe, but making it so requires some thought.
+            // See https://issues.jenkins-ci.org/browse/JENKINS-53724
             EnvVars.masterEnvVars.clear();
             EnvVars.masterEnvVars.putAll(enVars); 
         } catch (IllegalAccessException | NoSuchFieldException | NoSuchMethodException | InvocationTargetException exception) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectMasterEnvVarsSetter.java
```diff
@@ -3,19 +3,27 @@
...
  * @author Gregory Boissinot
  */
 public class EnvInjectMasterEnvVarsSetter extends MasterToSlaveCallable<Void, EnvInjectException> {
 
     private @Nonnull EnvVars enVars;
+    
+    // TODO: this may not be the version of core that gets the fix allowing us to
+    // use safe reflection to set the Platform. The exact version needs to be updated 
+    // after that version is released.
+    private static VersionNumber jenkinsVersionWithSetPlatform = new VersionNumber("2.139");
```
oleg-nenashev: `private static final`

src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectMasterEnvVarsSetter.java
```diff
@@ -24,22 +32,23 @@ public EnvInjectMasterEnvVarsSetter(@Nonnull EnvVars enVars) {
...
-            modifiersField.setAccessible(true);
-            modifiersField.setInt(masterEnvVarsFiled, masterEnvVarsFiled.getModifiers() & ~Modifier.FINAL);
-            masterEnvVarsFiled.set(null, enVars);
-        } catch (IllegalAccessException iae) {
-            throw new EnvInjectException(iae);
-        } catch (NoSuchFieldException nsfe) {
-            throw new EnvInjectException(nsfe);
+            
+            EnvVars.masterEnvVars.clear();
+            EnvVars.masterEnvVars.putAll(enVars);
```
oleg-nenashev: It is generally a terrible idea in Jenkins introduced before 2008.
The implementation here actually causes a concurrency risk since `masterEnvVars` is not synchronized. It may lead to concurrent modification exception
I would suggest replacing one collection by another instance instead of `clear()`=>`putAll()`

PR #133, opened 2018-10-25
[JENKINS-39403] - Automatically escape single backslashes in the properties loader

alexanderrtaylor:
Windows paths can now be used without the need to change all properties files to have the escaped \\ character

There is some funkyness that I had to regex for inside of the code in order to have /n work as intended. I just ignored all /n and //n that happens to ensure that the previous behavior is not touched. Maybe we should change that eventually but I wanted to focus my efforts on arbitrary / characters

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/envinject/service/PropertiesLoader.java
+++ b/src/main/java/org/jenkinsci/plugins/envinject/service/PropertiesLoader.java
@@ -71,7 +71,7 @@ public class PropertiesLoader implements Serializable {
 
         // Replace single backslashes with double ones so they won't be removed by Property.load()
         String escapedContent = content;
-        escapedContent = escapedContent.replaceAll("(?<=[^\\\\])\\\\(?=[^n])(?=[^\\\\])(?=[^\n])", "\\\\\\\\");
+        escapedContent = escapedContent.replaceAll("(?<=[^\\\\])\\\\(?![n|:|*|?|\"|<|>|\\||\\/])(?![\\\\])(?![\n])", "\\\\\\\\");
         Map<String, String> result = new LinkedHashMap<String, String>();
         StringReader stringReader = new StringReader(escapedContent);
         Properties properties = new Properties();
--- a/src/main/java/org/jenkinsci/plugins/envinject/service/PropertiesLoader.java
+++ b/src/main/java/org/jenkinsci/plugins/envinject/service/PropertiesLoader.java
@@ -68,8 +68,12 @@ public class PropertiesLoader implements Serializable {
     @Nonnull
     private Map<String, String> getVars(@Nonnull String content, @Nonnull Map<String, String> currentEnvVars) 
             throws EnvInjectException {
+
+        // Replace single backslashes with double ones so they won't be removed by Property.load()
+        String escapedContent = content;
+        escapedContent = escapedContent.replaceAll("(?<=[^\\\\])\\\\(?=[^n])(?=[^\\\\])(?=[^\n])", "\\\\\\\\");
         Map<String, String> result = new LinkedHashMap<String, String>();
-        StringReader stringReader = new StringReader(content);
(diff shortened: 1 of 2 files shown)


alexanderrtaylor:
@reviewbybees

src/test/java/org/jenkinsci/plugins/envinject/sevice/PropertiesLoaderTest.java
```diff
@@ -148,7 +148,7 @@ public void contentWithVarsToResolve() throws Exception {
     }
 
     private void checkWithVarsToResolve(boolean fromFile) throws Exception {
-        String content = "KEY1 =${VAR1_TO_RESOLVE}\nKEY2=https\\://github.com\nKEY3=${VAR3_TO_RESOLVE}\\\\otherContent";
+        String content = "KEY1 =${VAR1_TO_RESOLVE}\nKEY2=https://github.com\nKEY3=${VAR3_TO_RESOLVE}\\\\otherContent";
```
oleg-nenashev: This change makes me think that there are some implied regressions here

alexanderrtaylor: I misread this the first time. Why would you expect a variable `KEY2` which consists of `https:\\github.com` to resolve to `https://github.com`?

Those would be 2 different slash values and I would not expect that behavior. The test fails even without my changes

src/main/java/org/jenkinsci/plugins/envinject/service/PropertiesLoader.java
```diff
@@ -68,8 +68,12 @@
     @Nonnull
     private Map<String, String> getVars(@Nonnull String content, @Nonnull Map<String, String> currentEnvVars) 
             throws EnvInjectException {
+
+        // Replace single backslashes with double ones so they won't be removed by Property.load()
+        String escapedContent = content;
+        escapedContent = escapedContent.replaceAll("(?<=[^\\\\])\\\\(?=[^n])(?=[^\\\\])(?=[^\n])", "\\\\\\\\");
```
res0nance:
```suggestion
        escapedContent = escapedContent.replaceAll("(?<=[^\\\\])\\\\(?=[^n|\\W])(?=[^\\\\])(?=[^\n])", "\\\\\\\\");
```
If you do this you would not have to change the github example because that will cause a backwards incompatibility

src/test/java/org/jenkinsci/plugins/envinject/sevice/PropertiesLoaderTest.java
res0nance: @alexanderrtaylor I think @oleg-nenashev is concerned of the backwards compatibility

res0nance: Let me know if you can't work on this I'm blocked from using the latest plugin because of this very issue

src/main/java/org/jenkinsci/plugins/envinject/service/PropertiesLoader.java
res0nance: The negative lookahead will fix trailing backslash

res0nance: Actually looking at it again it might cause issues so i reverted that bit

src/test/java/org/jenkinsci/plugins/envinject/sevice/PropertiesLoaderTest.java
alexanderrtaylor: I am taking a look at your suggestion now and will try it out locally

alexanderrtaylor:
@oleg-nenashev fixed based on a suggestion from @res0nance who is a huge help here

The test change was reverted and it should not cause any regressions

oleg-nenashev:
I will integrate it into the next release

PR #134, opened 2018-12-28
Anonymous mastertoslavecallable

jsoref:
https://issues.jenkins-ci.org/browse/JENKINS-52946


pom.xml
```diff
@@ -90,6 +90,12 @@
     </build>
 
     <dependencies>
+        <dependency>
+            <groupId>org.jenkins-ci.lib</groupId>
+            <artifactId>envinject-lib</artifactId>
+            <version>1.27</version>
+        </dependency>
```
olivergondza: Why do we need to add this for this PR. To the best of my knowledge, the dependency was [recently replaced](https://github.com/jenkinsci/envinject-plugin/commit/af113cfd5008e73b349266891b89d6b043296816) with `envinject-api`.

olivergondza: This PR does not contribute any and master appears to be working without it. Putting aside it should be made a test-scoped dependency if this is truly needed for a test.

jsoref:
@olivergondza: fwiw, I dropped that change

jakub-bochenski:
Since the [plugin is up for adoption.](https://wiki.jenkins.io/display/JENKINS/EnvInject+Plugin) there might be nobody with write access to merge this.

olivergondza:
You are not wrong, the plugin is neglected and needs a proper maintainer. For now, some of us can help a bit to keep things from stopping.

oleg-nenashev:
Yes. I was de-facto maintaining it for 2 years while it was waiting for adoption, but I stopped doing it after the JENKINS-47370 mess (see my comment [here](https://issues.jenkins-ci.org/browse/JENKINS-47370?focusedCommentId=331761&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-331761)).

Anyone is welcome to take temporary or permanent ownership.

PR #138, opened 2019-03-06
[JENKINS-56449] - Add support for global config via configuration-as-code plugin

caitlin615:
[JENKINS-56449](https://issues.jenkins-ci.org/browse/JENKINS-56449)

Based on https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/docs/PLUGINS.md, adds support for defining EnvInject plugin global configuration within the configuration-as-code plugin.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -141,7 +141,7 @@
         <dependency>
             <groupId>io.jenkins</groupId>
             <artifactId>configuration-as-code</artifactId>
-            <version>1.7</version>
+            <version>1.23</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -57,7 +57,7 @@
     <properties>
         <ivy.plugin.version>1.21</ivy.plugin.version>
         <mockito.version>1.10.19</mockito.version>
-        <jenkins.version>1.625.3</jenkins.version>
+        <jenkins.version>2.14</jenkins.version>
         <java.level>7</java.level>
         <!-- Remove once JENKINS-45055 is fixed -->
         <findbugs.failOnError>false</findbugs.failOnError>
@@ -156,6 +156,12 @@
             <groupId>org.jenkins-ci.plugins.icon-shim</groupId>
             <artifactId>icon-shim</artifactId>
             <version>2.0.3</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.jenkins</groupId>
+            <artifactId>configuration-as-code</artifactId>
+            <version>1.7</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
+++ b/src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
@@ -35,6 +35,7 @@ import jenkins.model.GlobalConfigurationCategory;
 import jenkins.model.Jenkins;
 import net.sf.json.JSONObject;
 import org.jenkinsci.lib.envinject.EnvInjectException;
+import org.jenkinsci.Symbol;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.DataBoundConstructor;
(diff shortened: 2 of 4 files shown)


oleg-nenashev:
@caitlin615 could you please create a JIRA ticket and mark it with a `jcasc-compatibility` label? It is needed for the compatibility dashboard

caitlin615:
No problem! Created and updated the PR description

pom.xml
```diff
@@ -57,7 +57,7 @@
     <properties>
         <ivy.plugin.version>1.21</ivy.plugin.version>
         <mockito.version>1.10.19</mockito.version>
-        <jenkins.version>1.625.3</jenkins.version>
+        <jenkins.version>2.14</jenkins.version>
```
oleg-nenashev: It won't work properly. JCasC requires 2.60.3 at least. It is fine to bump the dependency (and to set java.level to 8)

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
```diff
@@ -65,6 +66,7 @@
 
     @Restricted(NoExternalUse.class)
     @RestrictedSince("2.0")
+    @DataBoundConstructor
```
oleg-nenashev: I doubt JCasC will alway take restricted constructors into account. CC @timja @casz

timja: hmm yeah probably worth adding a test to show that what you expect is exported:
https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/docs/PLUGINS.md#model-export-test

(also a different way of testing it in this PR here using a fixture instead of CNode):
https://github.com/jenkinsci/configuration-as-code-plugin/pull/860

oleg-nenashev:
#141 should make testing with JCasC possible.
@caitlin615 are you still interested in it?

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
oleg-nenashev: I just removed the restriction so that there should be no problem

oleg-nenashev:
Should we add a `Symbol` there, e.g. `envinject` or `envInjectSecurity`? The second name is pretty bad, but it will prevent us from suffering if non-security settings are added later. Not that I am afraid about it much, `envinject` is fine with me

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfiguration.java
```diff
@@ -46,6 +48,7 @@
  * @since 1.92
  */
 @Extension
+@Symbol("envinject")
```
timja: does this read better?
```suggestion
@Symbol("envInject")
```
(jcasc will resolve it case insensitively but it'll use the case on the symbol in the export)

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectPluginConfigurationTest.java
```diff
@@ -46,4 +47,14 @@ private void testRoundtrip(EnvInjectPluginConfiguration config) throws IOExcepti
         assertEquals("Value of enableLoadingFromMaster differs after the reload",
                 config.isEnableLoadingFromMaster(), reloaded.isEnableLoadingFromMaster());
     }
+
+    @Test
+    public void testConfigAsCode() throws Exception {
```
timja: please use: `@ConfiguredWithCode("configuration-as-code.yml")`
https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/docs/PLUGINS.md#configuration-test

would be good to do a full export test, (I would move this to a different class for just testing jcasc and then use a class rule to do an import and export test)
```suggestion
    @ConfiguredWithCode("configuration-as-code.yml")
    public void testConfigAsCode() throws Exception {
```

(you'll need to add an import for ConfiguredWithCode and add the tests classifier dependency (https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/docs/PLUGINS.md#how-to-test)

oleg-nenashev: Yeah. I was just too lazy to fix that + I did not want to modify a rule for all tests in the class.
Will kepp as is, but we can definitely refactor it later

PR #139, opened 2019-05-28
[JENKINS-49154] - Resolve job run by BUILD_NUMBER string value instead of the Integer conversion

wwojdanowski:
…n EnvInjectEnvVars


oleg-nenashev:
Thanks @wwojdanowski ! Nice to have more tests for this feature

PR #141, opened 2019-07-21
Update Jenkins core dependency to 2.138.4 + plugin POM facelift

oleg-nenashev:
I would like to apply some patches and to ensure that the new version builds with Java 11


oleg-nenashev:
Will go ahead to enable JCasC compatibility fixes

PR #143, opened 2019-09-01
Fix anonymous class serialization warning in EnvInjectComputerListener 

darxriggs:
The warning from org.jenkinsci.remoting.util.AnonymousClassWarnings was:
> WARNING: Attempt to (de-)serialize anonymous class org.jenkinsci.plugins.envinject.EnvInjectComputerListener$1
> see: https://jenkins.io/redirect/serialization-of-anonymous-classes/

It seems this anonymous class was missed in #134 @jsoref.

Hopefully this is the last fix in this regard. At least for `Callable` it should be fine now (`grep -rP 'new.*Callable.*{' src/` does not find anything else).

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectMasterEnvVarsRetriever.java
+++ b/src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectMasterEnvVarsRetriever.java
@@ -1,0 +1,17 @@
+package org.jenkinsci.plugins.envinject.service;
+
+import hudson.EnvVars;
+import jenkins.security.MasterToSlaveCallable;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+
+import java.io.IOException;
+import java.util.Map;
+
+@Restricted(NoExternalUse.class)
+public class EnvInjectMasterEnvVarsRetriever extends MasterToSlaveCallable<Map<String, String>, IOException> {
+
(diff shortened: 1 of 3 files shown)


darxriggs:
@oleg-nenashev can you have a look at this?

src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectMasterEnvVarsRetriever.java
```diff
@@ -0,0 +1,14 @@
+package org.jenkinsci.plugins.envinject.service;
+
+import hudson.EnvVars;
+import jenkins.security.MasterToSlaveCallable;
+
+import java.io.IOException;
+import java.util.Map;
+
+public class EnvInjectMasterEnvVarsRetriever extends MasterToSlaveCallable<Map<String, String>, IOException> {
```
oleg-nenashev: Add `@Restricted(NoExternalUse.class)`?

darxriggs: I already thought about this. But as most other classes in this package didn't use this annotation I did not add it for consistency reasons.

I'll just added it now.

oleg-nenashev:
Thanks @darxriggs !

oleg-nenashev:
Snipped it in 2.2.1. https://github.com/jenkinsci/envinject-plugin/releases/tag/envinject-2.2.1

PR #144, opened 2019-09-01
Code cleanup

darxriggs:
These are some minor code improvements.


src/main/java/org/jenkinsci/plugins/envinject/EnvInjectPasswordWrapper.java
```diff
@@ -197,7 +196,7 @@ public void makeSensitiveBuildVariables(AbstractBuild build, Set<String> sensiti
     /**
      * Class took from the mask-passwords plugin
      */
-    class EnvInjectPasswordsOutputStream extends LineTransformationOutputStream {
+    static class EnvInjectPasswordsOutputStream extends LineTransformationOutputStream {
```
oleg-nenashev: Makes sense to me though it might cause serialization issues in edge cases

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectEvaluatedGroovyScriptTest.java
```diff
@@ -186,7 +186,7 @@ public void testWorkaroundSecurity86() throws Exception {
         job.addProperty(property);
 
         TestUtils.saveConfigurationAs(jenkins, job, "alice");
-        //Since alice is an admin the script should be approved automagically
+        //Since alice is an admin the script should be approved automatically
```
oleg-nenashev: It was intended :)

PR #145, opened 2019-09-03
[JENKINS-39403] - Add test and simple cleanup

res0nance:
Follow up to #133 to add a test to prevent this functionality from breaking unknowingly.


oleg-nenashev:
I wonder why you needed to change the expression here. Any issues with the previous one?

res0nance:
They are equivalent instead of positive lookbehind with a negation i used the logical equivalent negative lookbehind. The vertical bars can also be omitted as they are sets of single characters.

PR #146, opened 2019-10-04
Escape paths for network shares

res0nance:
For windows paths #133 misses out the case of network shares where network shares begin with a double backslash. This PR adds that capability and a functional test for it.


res0nance:
@oleg-nenashev Could this be released?

oleg-nenashev:
OK, I will take a look.
This plugin is up for adoption, and I did not have much time to follow-up
on it, sorry

PR #149, opened 2019-10-07
[JENKINS-45055] Fix all spotbugs

res0nance:
Fix all remaining spotbugs in envinject

### DIFF
--- a/src/main/java/hudson/plugins/setenv/SetEnvBuildWrapper.java
+++ b/src/main/java/hudson/plugins/setenv/SetEnvBuildWrapper.java
@@ -49,7 +49,7 @@ public class SetEnvBuildWrapper extends EnvInjectMigrationBuildWrapper {
     public static final class DescriptorImpl extends BuildWrapperDescriptor {
         @Override
         public String getDisplayName() {
-            return null;
+            return "SetEnvBuildWrapper";
         }
 
         @Override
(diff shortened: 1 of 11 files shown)


res0nance:
I'm kinda stumped that this broke a test and because my machine refuses to run the tests I'm pretty much stuck

res0nance:
Apparently the older version of script security just doesn't run right on my system and updating made the test run :/

src/main/java/org/jenkinsci/plugins/envinject/EnvInjectJobPropertyInfo.java
```diff
@@ -10,18 +10,22 @@
...
  */
 public class EnvInjectJobPropertyInfo extends EnvInjectInfo implements Describable<EnvInjectJobPropertyInfo> {
 
+    private static final long serialVersionUID = 1L;
     @CheckForNull
     private final String scriptFilePath;
     @CheckForNull
     private final String scriptContent;
     @CheckForNull
+    @SuppressFBWarnings(value = "SE_BAD_FIELD", justification = "EnvInjectEvaluatedGroovyScriptTest#testWorkaroundSecurity86 relies on the serialization of this object")
```
res0nance: I'm not sure why we can't serialize this bit without Spotbugs complaining

src/main/java/hudson/plugins/setenv/SetEnvBuildWrapper.java
```diff
@@ -49,7 +49,7 @@ public Environment setUp(AbstractBuild build, Launcher launcher, BuildListener l
     public static final class DescriptorImpl extends BuildWrapperDescriptor {
         @Override
         public String getDisplayName() {
-            return null;
+            return "SetEnvBuildWrapper";
```
oleg-nenashev: If you put a name, these wrappers will start appearing in the lists IIRC

res0nance: Hmm i guess i could set it to a blank name so spotbugs wont complain, would that be ok?

oleg-nenashev:
Encoding and display name changes may cause regressions. I still need to review it carefully, not including into the today;s release.

PR #150, opened 2019-10-21
Move documentation from Wiki to GitHub

oleg-nenashev:
CC @res0nance , I would like to do this migration to cut a new release.

- [x] Convert Wiki to GitHub, move changelog to archive
- [x] Some minor documentation cleanup. Looks like some sections are still obsolete, but better than nothing


oleg-nenashev:
CI does not provide coverage for it anyway

PR #153, opened 2020-08-15
JENKINS-60559: Log warning when not all environment variables can be injected

rgroothuijsen:
When similarly-named environment variables with different case are defined as job properties, they will be treated as one variable and overwrite each other. `EnvVars` appears to treat variables in a case-insensitive manner by design, so if altering the existing behavior is not an option, a warning regarding this behavior should at least be logged to prevent confusion.


oleg-nenashev (changes requested):
This diagnostics message is definitely good. The problem is that the message goes to the system log, and there is no way to pin-point the root cause job from the message. I suggest either sending the message to TaskListener or adding more log details

oleg-nenashev (approved):

dhs-rec:
@oleg-nenashev, no, without telling the user which variable names actually clash, the message is NOT good. Sorry.

PR #154, opened 2020-12-07
Replace table/tr/td HTML tags by jelly constructs and div tags, and…

jimklimov:
…bump dependency versions to match

### DIFF
--- a/src/main/resources/util/hetero-list-readonly.jelly
+++ b/src/main/resources/util/hetero-list-readonly.jelly
@@ -57,23 +57,23 @@ THE SOFTWARE.
 
     <d:taglib uri="local">
         <d:tag name="body">
-            <table style="width:100%">
+            <div style="width:100%">
                 <j:set var="help" value="${descriptor.helpFile}"/>
                 <j:if test="${hasHeader}">
-                    <tr>
-                        <td colspan="3">
+                    <div style="tr">
+                        <f:entry>
                             <div class="dd-handle">
                                 <b>${descriptor.displayName}</b>
                             </div>
-                        </td>
+                        </f:entry>
                         <j:if test="${help!=null}">
-                            <td>
+                            <f:entry>
                                 <a href="#" class="help-button" helpURL="${rootURL}${help}">
                                     <img src="${imagesURL}/16x16/help.gif" alt="[help]" height="16" width="16"/>
                                 </a>
-                            </td>
+                            </f:entry>
                         </j:if>
-                    </tr>
+                    </div>
                     <!-- TODO: help support is unintuitive; people should be able to see help from drop-down menu -->
                     <j:if test="${help!=null}">
                         <f:helpArea/>
@@ -88,7 +88,7 @@ THE SOFTWARE.
                         <!--<f:repeatableDeleteButton value="${attrs.deleteCaption}" />-->
                     </div>
                 </f:block>
-            </table>
+            </div>
             <input type="hidden" name="kind" value="${descriptor.class.name}"/>
         </d:tag>
     </d:taglib>
--- a/src/main/resources/util/hetero-list-readonly.jelly
+++ b/src/main/resources/util/hetero-list-readonly.jelly
@@ -26,10 +26,14 @@ THE SOFTWARE.
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" xmlns:l="/lib/layout"
          xmlns:t="/lib/hudson" xmlns:f="/lib/form" xmlns:local="local">
     <st:documentation>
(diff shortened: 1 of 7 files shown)


jimklimov:
For reasons beyond my understanding, the self-tests fail with Jenkins code 2.269 (current) at org.jenkinsci.plugins.envinject.EnvInjectVarListTest.envVarsShouldBeCachedProperlyAfterReload(EnvInjectVarListTest.java:46) with `java.lang.AssertionError: TEXT_VAR is not present in the list`. The same test passes well with original old version constraint, and with 2.264 which should suffice for the table-vs-div problem (this one timed out once for me at org.jenkinsci.plugins.envinject.EnvInjectJobPropertyInfoTest.shouldNotCreateSecureGroovyScriptForNullScriptWithOldAPI with `Time elapsed: 180.332 s`).

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectVarListTest.java
```diff
@@ -16,44 +16,46 @@
...
         final Run<?, ?> reloadedBuild = p.getBuildByNumber(build.getNumber());
         {
             EnvInjectVarList varList = getVarListOrFail(reloadedBuild);
+            Assert.assertNotNull("EnvInject varList is null for the run after the reload", varList);
+
             Map<String, String> envMap = varList.getEnvMap();
-            Assert.assertNotNull("EnvInject vars list is null for the run after the reload", envMap);
+            Assert.assertNotNull("EnvInject envMap from varList is null for the run after the reload", envMap);
             Assert.assertTrue("TEXT_VAR is not present in the list", envMap.containsKey("TEXT_VAR"));
             Assert.assertEquals("TEXT_VAR has wrong value", "tvalue", envMap.get("TEXT_VAR"));
```
jimklimov, line 49: The commented problem about test that broke between 2.264 and 2.269 is here.

src/main/resources/util/hetero-list-readonly.jelly
```diff
@@ -57,23 +57,23 @@ THE SOFTWARE.
 
     <d:taglib uri="local">
         <d:tag name="body">
-            <table style="width:100%">
+            <div style="width:100%">
```
timja, line 60: This file should probably be rebased against Jenkins core, forks of core widgets are a pain...

jimklimov:
Aint my fault!

````
EC2 (aws) - Ubuntu 18.04 LTS (i-00a4410b6b57c97bb) was marked offline: Connection was broken: java.io.EOFException
````

src/main/resources/util/hetero-list-readonly.jelly
jimklimov: Do you mean there is a master copy of this file somewhere to "draw inspiration" from? :)

timja: yes here: https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/lib/form/hetero-list.jelly

src/test/java/org/jenkinsci/plugins/envinject/EnvInjectVarListTest.java
timja: most likely xstream changes https://www.jenkins.io/changelog/#v2.266

timja (commented):
@oleg-nenashev seems you were the last one to release here, you able to do one more or should I add myself?

oleg-nenashev:
I can cut the release. Would have preferred a compatibility layer, but IIUC
we have to deliver fixes insteas of keeping status quo

src/main/resources/util/hetero-list-readonly.jelly
jimklimov: Hi again, I took a look at this and have zero idea what the changes are about :-) 

I updated the file in this repo to `diff` with less noise against the one in upstream (modulo the change from 2-space indents to 4-space, and a few readability blank lines added in this one). But I think it would be beneficial if someone with half an idea about what that markup means does the actual transplants :)

oleg-nenashev: This extension point (EnvInjectJobPropertyContributor) is used in 2 plugins: shared objects (up for adoption, known vulnerabilities, 1000 installs) and ontrack (40 installs). Not sure maintaining that extension point worth any time spent

oleg-nenashev (approved):
I tested the new layout locally, and it seems to work. At the same time I am not sure the UX has actually improved after tables -> divs, the plugin ended up with a massive controls layout. It was not fancy before the change as well.

image

timja:
> I tested the new layout locally, and it seems to work. At the same time I am not sure the UX has actually improved after tables -> divs, the plugin ended up with a massive controls layout. It was not fancy before the change as well.
> 
> image

There's plans to iterate on it css wise, previously it wasn't possible to style properly and now we can.

PR #206, opened 2022-07-05
[JENKINS-68930] Fix Java17 Error on agent start

bulanovk:
See [JENKINS-66896](https://issues.jenkins.io/browse/JENKINS-66896) and [JENKINS-68930](https://issues.jenkins.io/browse/JENKINS-68930)

Use WA to work with modifiers field filtered since JDK12.
--add-opens java.base/java.lang=ALL-UNNAMED --add-opens java.base/java.lang.reflect=ALL-UNNAMED are required.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectMasterEnvVarsSetter.java
+++ b/src/main/java/org/jenkinsci/plugins/envinject/service/EnvInjectMasterEnvVarsSetter.java
@@ -7,7 +7,10 @@ import jenkins.security.MasterToSlaveCallable;
 import org.jenkinsci.lib.envinject.EnvInjectException;
 
 import java.lang.reflect.Field;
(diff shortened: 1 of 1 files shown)


NotMyFault:
Untested, but looks good so far. Thanks for filing the PR!

oleg-nenashev (changes requested):
Thanks for the fix! It will likely fail SpotBugs, because the return value of the new method might be null. In the end you should check for null and throw `NoSuchMethodException` to retain the behavior

bulanovk:
Any chance to get this PR merged?

NotMyFault (approved):
Looks good, do you mind taking another look @oleg-nenashev ?

limeman40:
Can someone review the change request so this could possibly be merged?

limeman40:
@oleg-nenashev Can you please review this so it can possible get merged?

PR #242, opened 2022-12-03
[JENKINS-69935] Set correct NODE_NAME on controller

MarkEWaite:
## [JENKINS-69935](https://issues.jenkins.io/browse/JENKINS-69935) Set correct NODE_NAME on controller

- Test NODE_NAME value on the controller
- [JENKINS-69935](https://issues.jenkins.io/browse/JENKINS-69935) Set correct NODE_NAME on controller

The `NODE_NAME` and `NODE_LABELS` environment variables defined by the EnvInject plugin were using `master` even after the migration to `built-in`.  Use the value from the controller instead of assuming it is always `master`.

Autoamated test confirms that it fails without the fix and passes with the fix.  Interactive testing performed with a freestyle job to confirm behavior of the new code is as expected.

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue


NotMyFault (approved):
Thanks for addressing!

